PR #4, opened 2024-12-10
feature: ctx module

rustatian:
# Reason for This PR

closes: https://jira.spiralscout.com/browse/EE2-1497

## Description of Changes

- Add context parser module

## API

In the root context (http component for example), initialize the `Contexter`:
```go
package foo
import (
	"context"

	ctxapi "github.com/ponyruntime/pony/api/context" // <--- module
	lengine "github.com/ponyruntime/pony/runtime/lua/engine"
	"go.uber.org/zap"
)

func foo() {
        log, _ := zap.NewDevelopment()
    	cter := ctxapi.NewContexter[any]() // contexter supports generics values, it can be `any` or might be typed with a structure or basic type.
        // store any value in it
        cter.WithValue("testKey", "testValue")
	cter.WithValue("testKey2", 1)
	cter.WithValue("testKey3", map[string]string{"test": "test"})

        // contexter might be extracted on any step before the ctx module:
        ctxer := ctx.Value(ctxapi.ContexterKey)
        // store additional values
        cter.WithValue("testKey222", 111)
        // save the contexter and pass to another function down the road
        ctx := context.WithValue(context.Background(), ctxapi.ContexterKey, cter)
        // as the last step, context with contexter should be passed to the lua engine constructor

       	engine := lengine.NewLuaEngine(ctx, log.Named("tests"))
}

``` 

Example of the Lua code:
```lua
	local ctx = require("ctx")
	local value, err = ctx.get("testKey")
	if err then
		return nil, err
	end
	return value
```
CC: @AndreiMaroz

### DIFF
--- a/api/context/context.go
+++ b/api/context/context.go
@@ -12,4 +12,24 @@ func (ck *Key) String() string {
 var (
 	LoggerKey      = &Key{name: "logger"}      //nolint:gochecknoglobals
 	CfgFilenameKey = &Key{name: "cfgfilename"} //nolint:gochecknoglobals
+	ContexterKey   = &Key{name: "contexter"}   //nolint:gochecknoglobals
 )
+
+type Contexter[T any] struct {
+	shared map[string]T
+}
+
+func NewContexter[T any]() *Contexter[T] {
+	return &Contexter[T]{
+		shared: make(map[string]T),
+	}
+}
+
+func (c *Contexter[T]) WithValue(key string, value T) {
+	c.shared[key] = value
+}
+
+func (c *Contexter[T]) Value(key string) (T, bool) {
+	v, ok := c.shared[key]
+	return v, ok
+}
--- a/runtime/lua/modules/ctx/ctx_test.go
+++ b/runtime/lua/modules/ctx/ctx_test.go
@@ -1,0 +1,70 @@
+package ctx
+
+import (
+	"context"
+	"testing"
+
+	ctxapi "github.com/ponyruntime/pony/api/context"
+	lengine "github.com/ponyruntime/pony/runtime/lua/engine"
+	"github.com/stretchr/testify/require"
+	"go.uber.org/zap"
+)
+
+func TestContextGetValue(t *testing.T) {
+	cter := ctxapi.NewContexter[any]()
+	cter.WithValue("testKey", "testValue")
+	cter.WithValue("testKey2", 1)
+	cter.WithValue("testKey3", map[string]string{"test": "test"})
+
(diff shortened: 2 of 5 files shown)


api/context/context.go
```diff
@@ -12,4 +12,24 @@ func (ck *Key) String() string {
 var (
 	LoggerKey      = &Key{name: "logger"}      //nolint:gochecknoglobals
 	CfgFilenameKey = &Key{name: "cfgfilename"} //nolint:gochecknoglobals
+	ContexterKey   = &Key{name: "contexter"}   //nolint:gochecknoglobals
```
wolfy-j, line 15: Nice, we can find a better name later. I would assume that semantically it's closer to values, since it's already part of Context object calling it contexted does not add much information.

runtime/lua/modules/ctx/ctx_test.go
```diff
@@ -0,0 +1,70 @@
...
+	"testing"
+
+	ctxapi "github.com/ponyruntime/pony/api/context"
+	lengine "github.com/ponyruntime/pony/runtime/lua/engine"
+	"github.com/stretchr/testify/require"
+	"go.uber.org/zap"
+)
+
+func TestContextGetValue(t *testing.T) {
+	cter := ctxapi.NewContexter[any]()
```
wolfy-j, line 14: nice

runtime/lua/modules/ctx/module.go
```diff
@@ -0,0 +1,62 @@
...
+
+import (
+	"github.com/ponyruntime/go-lua"
+	ctxapi "github.com/ponyruntime/pony/api/context"
+	"github.com/ponyruntime/pony/runtime/lua/engine"
+	"go.uber.org/zap"
+)
+
+// Module (ctx) accepts 1 argument, a string, and returns a context value found by that key.
+type Module[T any] struct {
```
wolfy-j, line 11: So this is the future common interface?

api/context/context.go
rustatian: Yeah, this is a tmp name.

wolfy-j (approved):

runtime/lua/modules/ctx/module.go
rustatian: This is just a Module (like env, sql, etc), not future. But it can be typed with other than `any` types.

rustatian: We can use here type unions if we want to, to restrict storing some not verified types.

PR #5, opened 2024-12-12
feature: EE2-1557 http handler module

rustatian:
# Reason for This PR

closes: https://jira.spiralscout.com/browse/EE2-1557

## Description of Changes

API: 
`Task` now contains two new methods (`Context` which returns the current context and `UpdateCtx` which updates (or set) current context for the `Task` + new field: `context`.)

New carrier type: (`types.go`): `HttpContextCarrier`
New context key for that: `HttpHandlerKey`

In the http layer, `HttpContextCarrier` should be initialized with the http request + response write and populated into the `Task`'s context via `UpdateCtx` method.

`http_handler` module contains a few basic methods: `method` which returns current http method + `write` which writes some data. This PR does not support multiple writes (because in Go, after the first write, StatusOK header would be sent automatically), but needed data structures have been prepared and located in the `http_wrapper.go` file (would be used via separate PR after API stabilization).


runtime/lua/modules/http_handler/http_wrapper.go
```diff
@@ -0,0 +1,119 @@
...
+}
+
+type lm struct {
+	pool sync.Pool
+	log  *zap.Logger
+}
+
+func (l *lm) Wrapper(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		wr := l.pool.Get().(*wrapper)
```
wolfy-j, line 97: can you explain this? does this module becomes multable?

PR #7, opened 2024-12-17
feature: vm-pool with ctx, update GHA, http_handler test with VM pool

rustatian:
# Reason for This PR

closes: https://jira.spiralscout.com/browse/EE2-1558

## Description of Changes

1. Update GHA, tests would be executed on every push.
2. Update internal vm-pool task with context.
3. Properly propogate context to the underlying vms and engines.


wolfy-j (approved):

go.mod
```diff
@@ -17,25 +17,19 @@ require (
...
-	github.com/xrash/smetrics v0.0.0-20240521201337-686a1a2994c1 // indirect
-	github.com/yuin/gopher-lua v1.1.1 // indirect
 	go.uber.org/multierr v1.11.0 // indirect
 	golang.org/x/net v0.32.0 // indirect
 	golang.org/x/sys v0.28.0 // indirect
 	golang.org/x/text v0.21.0 // indirect
-	google.golang.org/genproto/googleapis/rpc v0.0.0-20241206012308-a4fef0638583 // indirect
-	google.golang.org/protobuf v1.35.2 // indirect
-	gopkg.in/yaml.v3 v3.0.1 // indirect
+	google.golang.org/genproto/googleapis/rpc v0.0.0-20241216192217-9240e9c98484 // indirect
```
wolfy-j, line 33: do we still need to lock it to concrete version or it's more robust now?

go.sum
```diff
@@ -1,7 +1,11 @@
-github.com/cpuguy83/go-md2man/v2 v2.0.5 h1:ZtcqGrnekaHpVLArFSe4HK5DoKx1T0rq2DwVB0alcyc=
-github.com/cpuguy83/go-md2man/v2 v2.0.5/go.mod h1:tgQtvFlXSQOSOSIRvRPT7W67SCa46tRHOmNcaadrF8o=
 github.com/davecgh/go-spew v1.1.1 h1:vj9j/u1bqnvCEfJOwUhtlOARqs3+rkHYY13jYWTU97c=
 github.com/davecgh/go-spew v1.1.1/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
+github.com/go-logr/logr v1.4.2 h1:6pFjapn8bFcIbiKo3XT4j/BhANplGihG6tvd+8rYgrY=
```
wolfy-j, line 3: should we use it by default?

wolfy-j (approved):

runtime/lua/luapool/pool.go
```diff
@@ -102,9 +103,10 @@ func NewLuaPool(log *zap.Logger, scripts map[string]*Config, options ...Options)
 }
 
 // Queue adds task to the queue
-func (w *Pool) Queue(task *Task) <-chan string {
+func (w *Pool) Queue(ctx context.Context, task *Task) <-chan string {
```
wolfy-j, line 106: nice

wolfy-j: so we carry context separately from the task?

runtime/lua/modules/http_handler/http_handler_test.go
```diff
@@ -46,11 +45,62 @@ func TestInit(t *testing.T) {
 
 	time.Sleep(time.Second * 4)
 	httpserv.Close()
+	_ = resp.Body.Close()
+}
+
+func TestInitWithPool(t *testing.T) {
```
wolfy-j, line 51: perfect, how much options can we control in pool? can pool be "cold" a.k.a. compile on demand?

wolfy-j: ok i see config, in future it will be nice to have some simple "cold" start mechanism or LRU cache or something similar, but idea of putting each function into pool works perfectly, we have to make sure be able to reuse modules across calls, this way we can not only provide library access but also share data between pool

go.mod
rustatian: That is the temporal problem with their API. No need to do that while we don't use their API.

go.sum
rustatian: It's go.sum

runtime/lua/luapool/pool.go
rustatian: Yes, context might have some deadline mechanism in the Queue method. But atm it is just merged with the private Task context.

runtime/lua/modules/http_handler/http_handler_test.go
rustatian: > perfect, how much options can we control in pool?

Num VMs + num threads per function + functions we wanted to compile.

wolfy-j (approved):

PR #23, opened 2025-01-20
feature: EE2-1564 Make LFS context-specific

rustatian:
# Reason for This PR

ref: https://jira.spiralscout.com/browse/EE2-1564

## Description of Changes

1. Make LFS module context specific.
2. Detect absolute and relative paths. If we have a relative path, use saved current working directory + rel path, instead, use provided absolute dir.
3. New method: `close` which should be always called to close the FD for the stored in the user-data file.
4. All tests are green.
image

### DIFF
--- a/runtime/lua/modules/lfs/api.go
+++ b/runtime/lua/modules/lfs/api.go
@@ -2,48 +2,93 @@ package lfs
 
 import (
 	"os"
+	"path"
+	"path/filepath"
 	"time"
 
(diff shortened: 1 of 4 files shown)


wolfy-j (approved):
Will be nice to close file ref not on module level (this one is stateless) but using closer, see other modules that work with resources.

wolfy-j (approved):

PR #29, opened 2025-01-29
feature[EE2-1566]: native process code_executor

rustatian:
# Reason for This PR

ref: https://jira.spiralscout.com/browse/EE2-1566

## Description of Changes

- Update native code_executor according specs.
- lua specs updated.
- Separate stdout/stderr pipes.
- Fix potential problem with buffer reusage w/o copying.
- Add additional OK errors types (like file-already-closed) to the stream reader.
- Use NewResult instead of using structure directly (update code as well to eliminate &engine.Result{} -> engine.NewResult())


code_executors/native/native.go
```diff
@@ -29,17 +29,15 @@ type Executor struct {
 
 	cmd *exec.Cmd
 
-	stderrCh  chan []byte
-	stdoutCh  chan []byte
+	stderrp   io.ReadCloser
```
wolfy-j, line 34: gg

code_executors/native/native_test.go
```diff
@@ -64,8 +67,20 @@ func TestExecutor_MegaCommand(t *testing.T) {
 	}()
 
 	sb := new(strings.Builder)
-	for output := range executor.Stdout() {
-		sb.WriteString(string(output))
+	for {
+		// we don't care about the perf here
+		buf := make([]byte, 65536)
```
wolfy-j, line 74: does it make sense to expose buffer cusomization to user?

runtime/lua/engine/channel/layer.go
```diff
@@ -72,11 +72,11 @@ func (r *Layer) Send(ctx context.Context, ch *Channel, values ...lua.LValue) err
 				}
 
 				tg.Add(result.state)
-				err := tg.Send(ctx, engine.Result{
-					State:  result.state,
-					Result: result.values,
-					Error:  result.err,
-				})
+				err := tg.Send(ctx, engine.NewResult(
```
wolfy-j, line 75: OK makes sense

runtime/lua/modules/process/module.go
```diff
@@ -7,46 +7,19 @@ import (
 	"go.uber.org/zap"
 )
 
-const metatableName = "Process"
+const (
+	metatableName     = "proc"
```
wolfy-j: Process

runtime/lua/modules/process/module.go
```diff
@@ -7,46 +7,19 @@ import (
...
+	processStreamName = "process_stream"
```
wolfy-j: i dont think we need other stream name, default Stream will work

runtime/lua/modules/process/process.go
```diff
@@ -1,39 +1,85 @@
...
-
+	"github.com/ponyruntime/pony/code_executors/native"
 	"github.com/ponyruntime/pony/runtime/lua/modules/stream"
 	lua "github.com/yuin/gopher-lua"
+	"go.uber.org/zap"
 )
 
-func (m *Module) readProcess(l *lua.LState) int {
+func (m *Module) newProcess(l *lua.LState) int {
+	log := getCtxLogger(l)
```
wolfy-j, line 12: we prob want to move all modules to use this pattern, this way our ctx will contain correct leaf of logger since we can bind it any way we want in services layer

runtime/lua/modules/process/loader.go
```diff
@@ -23,8 +23,10 @@ func (m *Module) Loader(l *lua.LState) int {
 
 	mt := l.NewTypeMetatable(metatableName)
 	l.SetField(mt, "__index", l.SetFuncs(l.NewTable(), map[string]lua.LGFunction{
-		"start": m.startProcess,
-		"read":  m.readProcess,
+		"start":      m.startProcess,
+		"get_stderr": m.getStderr,
```
wolfy-j: stdout_stream, 
strderr_stream

wolfy-j: we need to register stream in closer

wolfy-j: stream(s)

runtime/lua/modules/process/process.go
```diff
@@ -1,39 +1,85 @@
...
+	l.Push(ud)
+	return 1
+}
+
+func (m *Module) getStdout(l *lua.LState) int {
+	log := getCtxLogger(l)
+	log.Debug("getting process stderr")
+	executor := getProcessExecutor(l)
+
+	s, errs := stream.NewStream(l.Context(), executor.StdoutReader(), stream.NewStreamConfig(65536))
```
wolfy-j: we should cache value i think as part of userdata of process, or maybe even create it ahead of time and register in closer

runtime/lua/modules/process/process.go
```diff
@@ -1,39 +1,85 @@
...
 	ud.Value = luaStream
 	l.SetMetatable(ud, l.GetTypeMetatable("Stream"))
 
 	l.Push(ud)
-	l.Push(lua.LNil)
-	return 2
+	return 1
+}
+
+func (m *Module) startProcess(l *lua.LState) int {
```
wolfy-j: so in Docker it will be the same semantics basically? like we can run remote process in isolated host? perfect

code_executors/native/native_test.go
rustatian: Good question. I do think that generally, user should not care about that. For fire-and-forget commands, 64k is a std way to handle IO. I think we should try 64k and see if we would have problems with that, leaving the space to quickly add options here to modify the buffer size.

runtime/lua/modules/process/module.go
rustatian: Fixed

runtime/lua/modules/process/module.go
rustatian: Yeah, it is not user actually, will remove it.

runtime/lua/modules/process/process.go
rustatian: Yeah, the same in Docker + we should add some option to let AI to choose an image.

runtime/lua/modules/process/loader.go
rustatian: done

runtime/lua/modules/process/process.go
rustatian: This is almost no-op operation. Caching might be done after the start operation, but let's see, no premature optimizations 😄

wolfy-j: it's less about that, it's more that we will have few streams pointing to the same byte stream and i'm not sure how it will work if it will, basically imagive ai asking for stream in a loop, it will be creating a lot of dummy wrappers at top of the same Reader

runtime/lua/modules/process/process.go
wolfy-j: ok, we will do it on registry level, not at code level, we will have to add code_executor manager that can register this stuff including configs, i'll explain how

wolfy-j (approved):

runtime/lua/modules/process/process.go
rustatian: Yeah, it makes sense to me.

PR #35, opened 2025-02-16
MVP

wolfy-j:
What it can do:
- processes, chats, background processing, data transformation
- llm agents and actors
- backend API and persistence
- more or less stable
- full spec prepared for ai

What is missing:
- temporal workflows
- llm funcs
- uploads (!)

Updates:
- added pubsub plan
- added pids
- added monitoring
- added process management
- hosts and nodes
- re-implemented btea and terminal as host instead of custom launcher
- added process service (supervisor)
- various api fixes
- ported all btea examples
- removed Tasks layer and ported everything to pubsub
- renamed couple of namespaces
- rewrote graph to use generics
- updated lua manager by splitting it into components
- added code manager and graph compiler for lua
- is is possible to insert funcs to other funcs now
- foundation for test builds
- added namespaces support (and refactored all components to use it)
- added the ability to declare multiple entries in a single file
- added ability to declare dependency on group or namespace
- added fully qualified name resolution for refencing ids in a name ns
- generated a bunch of new specs
- broke a lot of linters
- added clear semantics for cancel and etc
- sample apps
- updated func calling api
- func pubsub access
- added support for max process limiting per host
- added FS module and proper abstraction to isolate FS
- we can now server static content from FSs
- print now uses proper log
- replaced Closer with UoW that also can carry thread specific context values like curent dir
- added ability for process to cancel other processes
- added process inbox (potentially to replace custom listens)
- adds sql db resource management
- memory leaks hunting and destruction
- added demo demos

Getting close to wrapping this up.


wolfy-j:
todo: 
- fix linters and finalize tests
- multi process host +
- groups
- name registry +
- debug process exit in discord, make simpler example of process
- rollback needs more love
- workflow process (temporal + local host)
- links +
- migrations + (we can do real apps now, yey)

PR #49, opened 2025-04-15
pg driver changes

tuneyourserver:
# Reason for This PR

Change connection driver depends kind of db type connection

### DIFF
--- a/service/sql/factory.go
+++ b/service/sql/factory.go
@@ -36,11 +36,7 @@ func (f *DefaultPoolFactory) CreateStandardPool(kind registry.Kind, cfg *config.
 		return nil, fmt.Errorf("invalid connection config: %w", err)
 	}
 
-	driver := string(kind)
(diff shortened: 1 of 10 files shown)


service/sql/factory.go
```diff
@@ -36,7 +36,11 @@ func (f *DefaultPoolFactory) CreateStandardPool(kind registry.Kind, cfg *config.
 		return nil, fmt.Errorf("invalid connection config: %w", err)
 	}
 
-	db, err := sql.Open(string(kind), dsn)
+	driver := string(kind)
+	if kind == config.KindPostgres {
+		driver = "postgres"
```
wolfy-j: can you explain logic here?

tuneyourserver: db.sql.postgres is not a valid driver name for sql.Open

wolfy-j (approved):

PR #51, opened 2025-04-22
Feature/security

wolfy-j:
- Adds security layer


runtime/lua/security/access.go
```diff
@@ -0,0 +1,51 @@
...
+)
+
+const (
+	// STRICT determines if security checks are enforced when context is incomplete
+	// When false, incomplete security contexts will default to allow
+	STRICT = false
+)
+
+// Can checks if the action on the resource is allowed based on security context
+func Can(ctx context.Context, action, resource string, meta registry.Metadata) bool {
```
yramanovich: let's name it `Allow` or `IsAllowed`

yramanovich (approved):

PR #59, opened 2025-04-25
EE2-1713: add moduleloader

yramanovich:
# Reason for This PR

Integration with packer moduleloader.


system/registry/loader/entry_loader.go
```diff
@@ -24,18 +50,61 @@ type FileContent struct {
...
 
 	if content.Namespace == "" {
-		return nil, fmt.Errorf("namespace is required")
+		return nil, nil, fmt.Errorf("namespace is required")
+	}
+
+	newExports := make(map[string]Export, len(content.Exports))
+	entries := make([]registry.Entry, 0, len(content.RawEntries)+len(content.Exports))
+	for _, export := range content.Exports {
+		// TODO: make entries from exports
```
wolfy-j: ,

yramanovich: fixed

wolfy-j (approved):

api/registry/registry.go
```diff
@@ -90,6 +93,14 @@ type (
 		Meta Metadata `json:"meta"`
 		// Data is the actual payload associated with the entry
 		Data payload.Payload `json:"data"`
+		// Requirements list of parameters of required values.
+		Requirements []EntryRequirement `json:"requirements,omitempty"`
```
wolfy-j: actaully, make a proper payload sepatae from entry, is this only for ns.definition? than it should have own type

PR #61, opened 2025-04-29
[WIP] don't merge. EE2-1741 fix linter. intermediate commit

igordolgovspiralscout:
# Reason for This PR

update version of the linter and fix code issues 

## Description of Changes

linter updated, fixed linter issues, updated tests and changed CI/CD script


wolfy-j (approved):

api/process/context.go
```diff
@@ -11,12 +11,15 @@ import (
 // Context keys for storing process-related information in context.Context
 var (
 	// managerCtx is the context key for storing a process Manager.
+	//nolint:gochecknoglobals
```
yramanovich: I see it quite often - let's just remove this rule

runtime/lua/modules/hash/hash.go
```diff
@@ -197,7 +197,8 @@ func (m *Module) fnv64(l *lua.LState) int {
 	return 1
 }
 
-func (m *Module) hmacSha256(l *lua.LState) int {
+//nolint:revive
+func (m *Module) hmac_sha256(l *lua.LState) int {
```
yramanovich, line 201: Apply go naming here?

system/registry/registry_test.go
```diff
@@ -561,6 +562,7 @@ func TestInMemoryRegistry_Apply_Rollback_Failure(t *testing.T) {
 		},
 	}
 
+	//nolint:gocritic
```
yramanovich: Let's add rule https://golangci-lint.run/usage/linters/#nolintlint
require-explanation: true, so the reason is clear
```
//nolint:gocritic: explanation here

File: runtime/lua/engine/errors/errors_test.go [hunk 4]
```diff
@@ -226,7 +226,7 @@ func TestErrorIdentification(t *testing.T) {
 	}
 
 	// Build and print actual error chain for debugging
-	var actualChain []string
+	actualChain := make([]string, 0)
```
yramanovich: Do we really need to change it?
It changes semantics: from nil -> to allocated

runtime/lua/modules/hash/hash.go
igordolgovspiralscout: reverted

runtime/lua/engine/errors/errors_test.go
igordolgovspiralscout: agree. reverted as an empty map and nil map have different representation(at least for JSON value)

api/process/context.go
igordolgovspiralscout: removed

PR #63, opened 2025-04-30
Feature/ee2 1721 sql storage 1

tuneyourserver:
# Reason for This PR

sql ke value storage


wolfy-j (approved):

service/sqlstore/sqlstore.go
```diff
@@ -0,0 +1,475 @@
...
+			zap.String("error", err.Error()),
+			zap.String("resource", s.config.Database.Name))
+		return nil, err
+	}
+
+	db := conn.(sql.DBResource).DB
+
+	// Build query to retrieve value and check expiration
+	query := fmt.Sprintf(
+		"SELECT %s FROM %s WHERE %s = $1 AND (%s IS NULL OR %s > %s)",
```
igordolgovspiralscout, line 76: leads to SQL injection. we need to extend validation for config field to avoid it

api/service/sqlstore/config.go
```diff
@@ -0,0 +1,131 @@
...
+	// The store will run a background task at this interval to remove entries with expired TTLs
+	// Set to 0 to disable automatic cleanup
+	CleanupInterval time.Duration `json:"cleanup_interval"`
+
+	Lifecycle supervisor.LifecycleConfig `json:"lifecycle"`
+}
+
+// Validate checks if the configuration is valid
+// Returns an error if any configuration values are invalid
+func (c *SQLConfig) Validate() error {
```
igordolgovspiralscout, line 47: we need to tighten validation to avoid SQL injection

igordolgovspiralscout (changes requested):

wolfy-j:
Good to go, increase default check time interval.

igordolgovspiralscout (commented):
looks good. would be good to add table-name and database-name validation on `func (c *SQLConfig) Validate()`

igordolgovspiralscout (approved):

PR #67, opened 2025-05-05
EE2-1755 use ENV module to configure environment variables

igordolgovspiralscout:
# Reason for This PR

`EE2-1755 add env-system component for ENV variables`

## Description of Changes

`added a simple env-system component that allows defining ENV variable storage, and linking them to applications using a declarative YAML-based syntax`


service/envstorage/filestore/file.go
```diff
@@ -0,0 +1,84 @@
...
+	"sync"
+
+	"go.uber.org/zap"
+)
+
+// FileStorage implements envstorage.Storage interface using a JSON file
+type FileStorage struct {
+	path   string
+	values map[string]string
+	mutex  sync.RWMutex
```
rustatian: No need to use a regular map with a mutex. Use sync.Map and remove mutex. `defer` usage is not free. 
`sync.Map` is generally optimized for the case when entry is written once and read multiply times, which I suppose should be the case for the file storage. Or if multiply goroutines would write/overwrite/read the same file. This way you may significantly reduce lock contention.

service/envstorage/memstore/memory.go
```diff
@@ -0,0 +1,47 @@
...
+	"context"
+	"sync"
+
+	"go.uber.org/zap"
+)
+
+// MemoryStorage implements envstorage.Storage interface using in-memory map
+type MemoryStorage struct {
+	values map[string]string
+	mutex  sync.RWMutex
```
rustatian: Same as for the file storage.

wolfy-j:
Looks good so far.

api/env/env.go
```diff
@@ -0,0 +1,123 @@
...
+	StorageRegister event.Kind = "env.storageregister"
+	// StorageDelete    event.Kind = "env.storagedelete"
+	VariableRegister event.Kind = "env.variableregister"
+	VariableDelete   event.Kind = "env.variabledelete"
+	VariableUpdate   event.Kind = "env.variableupdate"
+)
+
+const (
+	KindVariable      registry.Kind = "env.variable"
+	KindStorageMemory registry.Kind = "env.storagememory"
```
wolfy-j, line 26: I would split actual implementatyions to own ns, it's servides while registry is system level, but it's OK for now.

app/src/demos/env_demo/_index.yaml
```diff
@@ -0,0 +1,103 @@
+version: "1.0"
+namespace: app.env.demo
+
+meta:
+  depends_on: [ "ns:system" ]
+  comment: "ENV Demo Application"
+
+entries:
+  - name: env_demo
+    kind: function.lua
```
wolfy-j, line 10: gg

app/src/demos/env_demo/env.lua
```diff
@@ -0,0 +1,65 @@
+local env = require("env")
+
+-- Main handler function
+local function handler()
+    local actual_api_key = nil
+    local actual_api_url = nil
+    local actual_document_loaders_host = nil
+
+    actual_api_key = env.get('openai_api_key')
```
wolfy-j: are you using entry name here? It should either be vbariable name (UPPERCASE) or ID, you can't use name without ns.

app/src/funcs/openai/index.yaml
```diff
@@ -30,4 +30,15 @@ entries:
...
\ No newline at end of file
+#      llm_query: app.funcs.openai:llm_query
+
+  - name: openai_api_key
+    kind: env.variable
+    meta:
+      type: secret_key
+      depends_on:
+        - envfile
+    variable: OPENAI_API_KEY
```
wolfy-j, line 41: this is public name

app/src/funcs/openai/index.yaml
```diff
@@ -30,4 +30,15 @@ entries:
 #    modules: [ "http_client", "json", "env", "time" ]
 #    imports:
 #      test: wippy.test:test
-#      llm_query: app.funcs.openai:llm_query
\ No newline at end of file
+#      llm_query: app.funcs.openai:llm_query
+
+  - name: openai_api_key
```
wolfy-j, line 35: Or this is public name wippy.llm.openai:open_api_key

wolfy-j (approved):

PR #75, opened 2025-05-26
EE2-1757 function interceptors

igordolgovspiralscout:
# Reason for This PR

`[Author TODO: add issue # or explain reasoning.]`

## Description of Changes

`[Author TODO: add description of changes.]`


api/interceptor/api.go
```diff
@@ -0,0 +1,41 @@
...
+import (
+	"context"
+	"net/http"
+
+	"github.com/ponyruntime/pony/api/function"
+	"github.com/ponyruntime/pony/api/runtime"
+)
+
+// Option defines a function that configures an interceptor
+type Option func(*Config)
```
wolfy-j: not quite sure what http does here

app/src/demos/interceptor_demo/_index.yaml
```diff
@@ -0,0 +1,67 @@
...
+      size: 2
+      workers: 4
+  - name: interceptor_demo.endpoint
+    kind: http.endpoint
+    meta:
+      comment: "interceptor demo endpoint"
+      router: system:api
+    method: GET
+    path: /interceptor/demo/otel
+    func: interceptor_demo_otel
```
wolfy-j, line 45: so you are passing context values to activate or deactivate interceptors or options?

api/interceptor/api.go
```diff
@@ -0,0 +1,41 @@
...
+func WithResponse(resp *http.Response) Option {
+	return func(c *Config) {
+		c.Response = resp
+	}
+}
+
+// Interceptor defines the interface for function execution interceptors
+type Interceptor interface {
+	// Handle processes the execution and calls next() to continue the chain
+	Handle(ctx context.Context, next func() *runtime.Result, opts ...Option) *runtime.Result
```
wolfy-j: we can use ctx activators instead of options. then you can allow http endpoint to share some config map (type Metadata for example) to allow to set Options of contex.

This way you can configure interceptors on pretty much any level and if you want you can make http endpoint with or without retry.

interceptors.md
```diff
@@ -0,0 +1,340 @@
...
+    StartTime    time.Time
+    EndTime      time.Time
+}
+```
+
+### Interceptor Interface
+
+```go
+type Interceptor interface {
+    Before(ctx context.Context, execution *Execution) error
```
wolfy-j: I assume it's old code.

runtime/lua/modules/funcs/module.go
```diff
@@ -51,11 +53,12 @@ func (m *Module) Name() string {
...
 	value.RegisterTypeMethods(l, "function.Executor", nil, map[string]lua.LGFunction{
-		"with_context": m.withContext,
-		"with_actor":   m.withActor,
-		"with_scope":   m.withScope,
-		"call":         m.call,
-		"async":        m.async,
+		"with_context":     m.withContext,
+		"with_actor":       m.withActor,
+		"with_scope":       m.withScope,
+		"with_interceptor": m.withInterceptor,
```
wolfy-j: with_options, functions dont know anything about interceptors really, they only have some calling context, actors, scopes and values and options, this options are the same as one you can configure on endpoint and they work as activators of interceptors (not order, but rather when interceptor decides if it has to do somthing or not).

system/interceptor/ratelimit.go
```diff
@@ -0,0 +1,74 @@
...
+	pid, ok := pubsub.GetPID(ctx)
+	if !ok {
+		// Handle case where PID is not found in context
+		return &runtime.Result{Error: fmt.Errorf("PID not found in context")}
+	}
+
+	pidStr := pid.String()
+
+	// Get or create rate limiter for this PID
+	limiter, ok := i.cache.Get(pidStr)
```
wolfy-j: not sure this is going to work really

api/function/funcs.go
```diff
@@ -25,6 +25,11 @@ const (
 	Accept event.Kind = "function.accept"
 	// Reject is sent FROM function nodes when a handler registration is rejected
 	Reject event.Kind = "function.reject"
+
+	OptionsRegister event.Kind = "function.optionsregister"
```
yramanovich, line 29:
```suggestion
	OptionsRegister event.Kind = "function.options.register"
```

api/interceptor/api.go
```diff
@@ -0,0 +1,65 @@
...
+	if err != nil {
+		return err
+	}
+	*d = Duration(dur)
+	return nil
+}
+
+func (d Duration) MarshalJSON() ([]byte, error) {
+	return json.Marshal(time.Duration(d).String())
+}
```
yramanovich: Could you explain why do we need this? I think duration serialisation works by default

api/interceptor/api.go
```diff
@@ -0,0 +1,65 @@
...
+
+type RetryOptions struct {
```
yramanovich, line 27: I would propose to store RetryPolicies here instead
```go
type RetryConfig struct {
    // Only one of these should be set. We can extend later
    Attempts       *AttemptsConfig       `json:"attempts,omitempty"`
    Backoff        *BackoffConfig        `json:"backoff,omitempty"`
}

type AttemptsConfig struct {
    Max int `json:"max"`
}

type BackoffConfig struct {
    Strategy string  `json:"strategy"`
    Initial  int     `json:"initialMs"`
    Max      int     `json:"maxMs"`
    Factor   float64 `json:"factor,omitempty"`
}
```

api/interceptor/context.go
```diff
@@ -0,0 +1,48 @@
...
+
+type CancelContextKey struct{}
+
+// WithInterceptor adds the interceptor registry to the context
+func WithInterceptor(ctx context.Context, registry Registry) context.Context {
+	return context.WithValue(ctx, RegistryContextKey{}, registry)
+}
+
+// GetInterceptors retrieves the interceptor registry from the context
+func GetInterceptors(ctx context.Context) Registry {
```
yramanovich, line 20: What if we return second parameter which indicates if value is found
`func GetInterceptors(ctx context.Context) (Registry, bool)`

I can see we don't use it in other places, but I think it makes sense as far we can really check interface for nil (only reflection)

api/interceptor/context.go
```diff
@@ -0,0 +1,48 @@
...
+	}
+	return nil
+}
+
+func GetOptionsFromContext(ctx context.Context) Options {
+	if options, ok := ctx.Value(OptionsContextKey{}).(Options); ok {
+		return options
+	}
+
+	return Options{}
```
yramanovich, line 32: I can see we return zero value here. Does it work in case RequestsPerSecond - 0?

cmd/runner/main.go
```diff
@@ -120,14 +123,18 @@ import (
 	"github.com/ponyruntime/pony/system/security"
 	"github.com/ponyruntime/pony/system/supervisor"
 	"github.com/ponyruntime/pony/system/topology"
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc"
+	otelresource "go.opentelemetry.io/otel/sdk/resource"
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+	semconv "go.opentelemetry.io/otel/semconv/v1.4.0"
```
yramanovich, line 134:
```suggestion
	semconv "go.opentelemetry.io/otel/semconv/v1.36.0"
```

cmd/runner/main.go
```diff
@@ -221,6 +230,7 @@ func NewApp(verbose, veryVerbose bool) (*App, error) {
 		topo:          topo,
 		pidReg:        pidReg,
 		forceShutdown: make(chan struct{}),
+		otelCleanup:   nil, // will be set in Start
```
yramanovich, line 237: Let's have noop func here

cmd/runner/main.go
```diff
@@ -712,9 +748,19 @@ func loadApplicationState(
 		vars[pair[0]] = pair[1]
 	}
 
+	// Initialize OpenTelemetry
+	cleanup, err := initOpenTelemetry(
+		context.Background(),
+		os.Getenv("OTEL_ENDPOINT"),
```
yramanovich, line 719: We don't need to that manually. otel packages automatically resolves it internally

cmd/runner/main.go
```diff
@@ -1033,3 +1079,76 @@ func WithLuaRuntime(a *App) []eventbus.EventHandler {
...
+	)
+	otel.SetTracerProvider(tp)
+
+	// Store cleanup function
+	cleanup := func() {
+		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+		defer cancel()
+
+		if err := tp.Shutdown(ctx); err != nil {
+			fmt.Printf("Error shutting down tracer provider: %v\n", err)
```
yramanovich: Let's use logger instead

cmd/runner/main.go
```diff
@@ -1033,3 +1079,76 @@ func WithLuaRuntime(a *App) []eventbus.EventHandler {
...
+		}
+	}
+
+	// Create and end a test span to verify tracing works
```
yramanovich: Don't forget to remvoe it

system/interceptor/manager.go
```diff
@@ -0,0 +1,128 @@
...
+
+	m.logger.Info("sent interceptor update request",
+		zap.String("id", entry.ID.String()))
+	return nil
+}
+
+// Delete implements registry.EntryListener
+func (m *Manager) Delete(_ context.Context, entry registry.Entry) error {
+	// Send delete event to the registry
+	m.eventBus.Send(context.Background(), event.Event{
```
yramanovich: Pass context correctly

system/interceptor/otel.go
```diff
@@ -0,0 +1,78 @@
...
+// NewOTelInterceptor creates a new OpenTelemetry interceptor
+func NewOTelInterceptor() *OTelInterceptor {
+	return &OTelInterceptor{
+		tracer: otel.Tracer("wippy"),
+	}
+}
+
+// Handle implements the interceptor interface
+func (i *OTelInterceptor) Handle(ctx context.Context, next func() *runtime.Result) *runtime.Result {
+	if i.tracer == nil {
```
yramanovich, line 27: I think we can omit this logic. Just use global trace-provider. If it's not initialized it's going to be noop.

system/interceptor/ratelimit.go
```diff
@@ -0,0 +1,92 @@
...
+	return next()
+}
+
+// Format implements the payload.Payload interface
+func (i *RateLimitInterceptor) Format() payload.Format {
+	return payload.Golang
+}
+
+// Data implements the payload.Payload interface
+func (i *RateLimitInterceptor) Data() any {
```
yramanovich: @wolfy-j isn't it should be convertible to different formats like json, etc.?

api/function/funcs.go
igordolgovspiralscout: @yramanovich main reason - eventbus.NewSubscriber() regex don't work with dots https://github.com/wippyai/runtime/blob/f316171758d7f424d62aa4751bf7279146ec99f9/system/function/functions.go#L55-L55

api/interceptor/api.go
igordolgovspiralscout: nice catch, that's redundant

api/interceptor/api.go
igordolgovspiralscout: sure, it might be extended later to support different strategies(fixed, backoff, conditional, circuit breaker, etc.)

api/interceptor/context.go
igordolgovspiralscout: good idea, but we are already using a similar approach for other "context" functions, for example:
https://github.com/wippyai/runtime/blob/f316171758d7f424d62aa4751bf7279146ec99f9/api/function/context.go#L20
 - since both approaches are essentially identical, I suggest sticking with the existing one

api/interceptor/context.go
igordolgovspiralscout: in this case, "retry" interceptor will be disabled https://github.com/wippyai/runtime/blob/f316171758d7f424d62aa4751bf7279146ec99f9/system/interceptor/retry.go#L30-L32

wolfy-j:
Please resolve pending conversations. Thank you

app/src/demos/interceptor_demo/_index.yaml
igordolgovspiralscout: optdated

cmd/runner/main.go
igordolgovspiralscout: optdated

api/interceptor/api.go
igordolgovspiralscout: added with_options support:
```
    local executor = funcs.new()
    local executor_with_options = executor:with_options({
        retry = {
            attempts = 10
        },
        ratelimit = {
            rps = 1,
            burst = 1
        },
        timeout = {
            timeout = "200ms"
        }
    })
```

system/interceptor/ratelimit.go
igordolgovspiralscout: @wolfy-j for now we support only `payload.Golang` https://github.com/wippyai/runtime/pull/75/files/0fc147c3f8dc796937d942185e83d85cf246b034#diff-dd56fe5969b4a6807757927aa3ff6a8e6e6fdf77e48d911b51890a807de8eff1R85-R87

igordolgovspiralscout: fixed

wolfy-j:
Ready to merge.

PR #79, opened 2025-06-04
added the ability to avoid https certificate verification

tuneyourserver:
# Reason for This PR

added the ability to avoid https certificate verification


wolfy-j:
LInters please

wolfy-j:
Also, what is the core issue so you have to bypass SSL on it? Can't you configure proper CA?

tuneyourserver:
Probably you're right. It doesn't make sense to make a flag in the main branch yet. I'll leave it in separated branch. Thanks

PR #101, opened 2025-07-31
EE2-1856: Improve module loading with selective entry loading

igordolgovspiralscout:
# Reason for This PR

`[Author TODO: add issue # or explain reasoning.]`
- Modified Manager.Load() to return LoadResult with loaded modules info
- Added loadEntriesFromLoadedModules() to load entries only from specific modules
- Updated app.go to use new LoadResult and selective entry loading
- Enhanced module loading flow to be more efficient and targeted
- Updated tests and manifest handling to support new loading approach


wolfy-j (approved):
seems ok to me

PR #115, opened 2025-09-04
feat(wippy): EE2-1877 add module requirement default

igordolgovspiralscout:
### Overview
Adds default value support for module requirements, making the `parameters` section in `ns.dependency` entries **optional** when modules provide sensible defaults.

### Key Changes
- **Default Value Support**: Module requirements can declare `default` values
- **Optional Parameters**: `parameters` section becomes optional when defaults exist
- **Graceful Fallback**: Handles missing/malformed parameters by falling back to defaults
- **Enhanced Validation**: Parameter-requirement matching with warning logs

### Implementation
- Added `resolveParameterValue()` method with dependency → default fallback logic
- Added `getRequirementDefaultValue()` and `validateParameterMatching()` helpers
- Enhanced `ResolveModuleDefinitions()` with default value support
- Comprehensive test suite (566+ new test lines)

### Usage Example
```yaml
# Module with default
- name: application_host
  kind: ns.requirement
  default: "app:processes"  # Makes parameters optional

# Application (simplified)
- name: __dependency.wippy.llm
  kind: "ns.dependency"
  component: "wippy/llm"
  # No parameters section needed!
```

### Benefits
- **Simplified Configuration**: Less boilerplate when using modules
- **Flexible Override**: Still can override defaults when needed
- **Backward Compatible**: Existing apps work unchanged
- **Better DX**: More intuitive module usage

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,7 @@ Wippy enables systems that can understand their own structure and adapt over tim
 *   **Concurrency Model:** Utilizes coroutines and Go-inspired channels for efficient, non-blocking concurrency within and between processes.
 *   **Command-Based Interface:** Modern CLI architecture with intuitive commands for dependency management and system operations.
 *   **Module Replacements:** Local development support through module replacement system for faster iteration cycles.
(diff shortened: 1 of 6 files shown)


roxblnfk (approved):
Works well

PR #127, opened 2025-09-30
feat(wippy): EE2-1876 update workflow to publish releases instead of commits

igordolgovspiralscout:
- Change triggers to only run on release tags (v*, release/*)
- Update binary names to include version tag (e.g., wippy-darwin-amd64-1.2.3-alpha)
- Replace wippy-releases commits with GitHub releases
- Remove PR workflow and simplify release process
- Publish binaries as assets in wippy-releases repository


igordolgovspiralscout:
moved to gitlab

PR #130, opened 2025-10-13
fix(wippy): EE2-1925 use temporary directory for wippy update to prevent module directory modification

igordolgovspiralscout:


wolfy-j (approved):

igordolgovspiralscout:
implemented at https://github.com/wippyai/runtime/pull/132

PR #135, opened 2025-11-05
fix(wippy): EE2-1954 fix permission condition to use exact match

igordolgovspiralscout:
Fix Permission Condition Evaluation for Role Arrays (EE2-1954)

Fixed bug in policy condition evaluator where contains/ncontains operators incorrectly performed substring matching on role arrays instead of exact element matching. Now actor.meta.roles ncontains "admin" correctly returns TRUE when roles are ["distributor-admin", "super-admin"].
Changes: Modified toString() to prevent array-to-string conversion + added comprehensive test suite

### DIFF
--- a/.cursorrules
+++ b/.cursorrules
@@ -1,0 +1,28 @@
+# Go Testing Rules
+
+## ✅ DO:
+- Use `testify`: `require.NoError(t, err)` and `assert.Equal(t, want, got)`
+- Always add `t.Parallel()` to tests and subtests
+- Use table-driven tests with descriptive names
+- Group tests with `t.Run("group name", func(t *testing.T) {...})`
+- Test edge cases: nil, empty, zero, boundary values
+- Extract repeated values to constants at top of test
+- Name tests: `TestType_Method_Scenario` format
+- Use `assert.Error(t, err)` or `assert.NoError(t, err)` for error checks
+
+## ❌ DON'T:
+- Don't use `if (err != nil) != tt.wantErr` - use `assert.Error` or `assert.NoError`
+- Don't add `wantErr` field if it's always the same value - just check explicitly
+- Don't use plain `if got != want` - use `assert.Equal`
+- Don't use `t.Fatalf` for non-critical checks - use `assert`
+- Don't skip `t.Parallel()` - tests must run in parallel
+- Don't use `t.Logf` for test progress - test names should be self-explanatory
+- Don't write comments explaining expected behavior - test name should be clear
+- Don't repeat test data - use constants or table-driven tests
+- Don't write code that takes 15 seconds to understand - keep it simple
+
+## Before Commit:
+```bash
+make lint  # Must be 0 issues
+```
+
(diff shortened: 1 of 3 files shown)


service/policy/condition_evaluator_test.go
```diff
@@ -1135,6 +1135,176 @@ func TestTypeConversions(t *testing.T) {
...
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := evaluator.EvaluateCondition(tt.condition, tt.actor, tt.action, tt.resource, tt.meta)
+			if (err != nil) != tt.wantErr {
+				t.Errorf("EvaluateCondition() error = %v, wantErr %v", err, tt.wantErr)
+				return
+			}
+			if got != tt.want {
+				t.Errorf("EvaluateCondition() got = %v, want %v\nComment: %s", got, tt.want, tt.comment)
+			} else {
+				t.Logf("✓ %s", tt.comment)
```
rustatian: After you generate a test with AI, please, re-read it and think. Why do you have this comment with an AI `✓` mark here on ok branch? What is the reason for me, who runs this test (for example), to see that on successful execution?

igordolgovspiralscout: removed t.Logf("✓ %s", tt.comment)

.cursorrules
```diff
@@ -0,0 +1,28 @@
+# Go Testing Rules
+
+## ✅ DO:
+- Use `testify`: `require.NoError(t, err)` and `assert.Equal(t, want, got)`
+- Always add `t.Parallel()` to tests and subtests
```
rustatian: Why always use t.Parallel()?

service/policy/condition_evaluator_test.go
```diff
@@ -1135,6 +1138,383 @@ func TestTypeConversions(t *testing.T) {
...
+		}
+	})
+
+	t.Run("contains operator on string fields", func(t *testing.T) {
+		t.Parallel()
+
+		tests := []struct {
+			name      string
+			field     string
+			value     interface{}
```
rustatian: any

service/policy/condition_evaluator_test.go
```diff
@@ -1135,6 +1138,383 @@ func TestTypeConversions(t *testing.T) {
...
+		testActorID  = "user123"
+	)
+
+	t.Run("contains operator on array fields", func(t *testing.T) {
+		t.Parallel()
+
+		tests := []struct {
+			name      string
+			field     string
+			value     interface{}
```
rustatian: 'any', tell your ai to use any instead of interface{}

service/policy/condition_evaluator_test.go
```diff
@@ -1135,6 +1138,383 @@ func TestTypeConversions(t *testing.T) {
...
+	)
+
+	t.Run("contains operator on array fields", func(t *testing.T) {
+		t.Parallel()
+
+		tests := []struct {
+			name      string
+			field     string
+			value     any
+			actorMeta registry.Metadata
```
rustatian, line 1163: As mentioned in Discord, for the tdt, use https://pkg.go.dev/github.com/stretchr/testify/assert#example-ErrorAssertionFunc to provide clear conditions on what the expectation and result.

igordolgovspiralscout: updated

rustatian (approved):

PR #136, opened 2025-11-06
fix(wippy): EE2-1956 avoid loading modules folder (as part of project files) when it’s located inside the project directory

igordolgovspiralscout:
### Changes

- Added filtering for `.wippy` and `replacements` directories during source scanning
- Added `--lock` flag to `update` command (updates lock file without installing)
- Removed `-m, --modules-dir` and `-d, --src-dir` flags (now use `.wippy` and `.` by default)
- Optimized `filteredFS` with pre-cleaned paths for better performance

### Breaking Changes

⚠️ Removed CLI flags: `-m, --modules-dir`, `-d, --src-dir`

### DIFF
--- a/cmd/runner/app/deps_test.go
+++ b/cmd/runner/app/deps_test.go
@@ -109,7 +109,6 @@ func TestDependencyManager_prepareExcludeDirs(t *testing.T) {
 	}
 
 	for _, tt := range tests {
-
 		t.Run(tt.name, func(t *testing.T) {
 			t.Parallel()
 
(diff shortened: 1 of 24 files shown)


cmd/runner/app/deps_test.go
```diff
@@ -0,0 +1,215 @@
+package app
```
rustatian, line 1: Update tests to use normal assert/require + https://pkg.go.dev/github.com/stretchr/testify/assert#example-ErrorAssertionFunc

wolfy-j (approved):

cmd/runner/app/deps_test.go
igordolgovspiralscout: fixed

rustatian (approved):

PR #140, opened 2025-11-18
fix(wippy): EE2-1983 fix wippy replace on windows

igordolgovspiralscout:
### Summary
Fixes module path resolution in `wippy replace` to work correctly on Windows. Refactors path resolution logic for cross-platform compatibility.

#### Core Fixes
- **`createProjectRootFS`**: Uses `filepath.Abs()` and `filepath.Clean()` instead of `os.Getwd()` to properly handle Windows paths
- **`resolveModulePath`**: Complete rewrite using `filepath.IsAbs()` and `filepath.Rel()` for cross-platform path handling
- **Base path resolution**: Uses lock file directory instead of current working directory as the base for resolving module paths
- **Module loading**: Handles modules outside project root by creating separate filesystems

#### Testing
  - Absolute and relative paths
  - Cross-platform path separators
  - Windows-specific scenarios
  - Edge cases (empty paths, multiple slashes, special characters)
- Fixed flaky test in `system/eventbus/bus_test.go`

### DIFF
--- a/cmd/runner/app/app.go
+++ b/cmd/runner/app/app.go
@@ -717,7 +717,7 @@ func (a *App) loadApplicationState(ctx context.Context, appFS iofs.FS) (regapi.C
 			return nil, nil, fmt.Errorf("parent dependency conflicts detected: %w", err)
 		}
 
-		dependencyEntries, err := loadEntriesFromLoadedModules(ctx, folderLoader, loadResult, projectRootFS, a.logger, parentDependencyMap)
+		dependencyEntries, err := loadEntriesFromLoadedModules(ctx, folderLoader, loadResult, projectRootFS, a.config.folderPath, a.logger, parentDependencyMap)
 		if err != nil {
 			return nil, nil, fmt.Errorf("load dependencies: %w", err)
 		}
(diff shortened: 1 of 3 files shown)


cmd/runner/app/app.go
```diff
@@ -901,28 +901,38 @@ func createProjectRootFS(_ string) (iofs.FS, error) {
...
-				zap.String("finalModulePath", resolvedPath))
+		// Convert absolute path to relative path from current working directory
+		relPath, err := filepath.Rel(currentDir, modulePath)
+		if err != nil {
+			return "", fmt.Errorf("failed to resolve relative path from %s to %s: %w", currentDir, modulePath, err)
+		}
 
-			return resolvedPath, nil
+		// Check if the path escapes from the current directory (security check)
+		if strings.HasPrefix(relPath, "..") {
```
rustatian: I'm not sure about this check. Usually, this security check is used for HTTP (or external) requests, when you use the PATH from it. Unverified sources. But here, if somebody has access to overwrite a lock or module path directly on a server (internal file system)... You have a much bigger problem than checking for `..` into `/etc/passwd`. So I think this is a pointless check.

igordolgovspiralscout: removed defense for path traversal attacks

wolfy-j (approved):

rustatian (approved):

wolfy-j:
This code is outdated, in se branch I have new implementation of lock file with proper composition, can you please check if this issue still exists there.

PR #238, opened 2026-04-02
Add network overlay layer for transparent traffic routing through

skhaz:
## Summary

- Transparent overlay network routing through **Tor**, **I2P**, and **Tailscale**
- Fixed DNS leak in SSRF protection (was resolving hostnames locally before overlay routing)
- Fixed I2P SAM v3 protocol: two-connection handshake + port stripped from DESTINATION field
- Triple-checked all 10 traffic paths — zero leaks

## Test Results — 69/69 PASS, 0 SKIP, 0 FAIL (`-race` enabled)

| Network | Unit | E2E | Total |
|---------|------|-----|-------|
| I2P | 17 | 4 | 21 |
| Tailscale | 21 | 8 | 29 |
| Tor | 13 | 6 | 19 |
| **Total** | **51** | **18** | **69** |


service/net/tor.go
```diff
@@ -0,0 +1,81 @@
...
+
+	// When stream isolation is enabled, create a per-connection SOCKS5 dialer
+	// with random credentials. Tor treats each unique user:pass pair as a
+	// separate circuit, providing stream isolation.
+	if s.isolateStreams {
+		cred, err := randomIsolationCredential()
+		if err != nil {
+			return nil, fmt.Errorf("tor: stream isolation credential: %w", err)
+		}
+		d, err := proxy.SOCKS5("tcp", s.addr, cred, proxy.Direct)
```
wolfy-j: is it just sock proxy for tor gateway?

service/net/helpers_test.go
```diff
@@ -0,0 +1,352 @@
...
+// tailscaleEnv returns the Tailscale auth key and control URL for E2E tests.
+//
+// Resolution order for auth key:
+//  1. TS_AUTHKEY environment variable (explicit key)
+//  2. Auto-generate from running Headscale Docker container
+//
+// When auto-generating, TS_CONTROL_URL defaults to http://localhost:8090
+// (the standard Headscale docker-compose port mapping).
+func tailscaleEnv() (authKey, controlURL string) {
+	authKey = os.Getenv("TS_AUTHKEY")
```
wolfy-j: need via proper wippy config or rewritable

wolfy-j (commented):
see comemnts

service/net/tailscale.go
```diff
@@ -0,0 +1,90 @@
...
+package net
+
+import (
+	"context"
+	"fmt"
+	"net"
+	"os"
+
+	netapi "github.com/wippyai/runtime/api/net"
+	"tailscale.com/tsnet"
```
wolfy-j: other services like tor looks like simple sock roixes, should we nuke them and only sock + tailscale to avoid shallow adapters?

api/runtime/lua/config.go
```diff
@@ -62,6 +62,7 @@ type (
 		Meta    attrs.Bag              `json:"meta,omitempty"`
 		Source  string                 `json:"source"`
 		Method  string                 `json:"method"`
+		Network string                 `json:"network,omitempty"`
```
wolfy-j: please use options from meta, trace how they merged it will give you runtime capabilities also

service/net/tor.go
skhaz: Yes, it's a SOCKS5 client dialer that connects through a Tor proxy (e.g. the Tor daemon's SOCKS port, typically 9050). The only extra logic is optional stream isolation — each `DialContext` call generates random SOCKS5 credentials so Tor assigns a separate circuit, preventing correlation between connections. DNS for .onion addresses is resolved remotely by the proxy.

service/net/helpers_test.go
skhaz: These are test helpers only (in `_test.go`), and all defaults are already overridable via env vars (`HEADSCALE_CONTAINER`, `HEADSCALE_USER`, `TS_AUTHKEY`, `TS_CONTROL_URL`, `TOR_PROXY_HOST`, etc.). Could you clarify what you mean by "proper wippy config" here — should the E2E tests load from a wippy config YAML file instead of env vars, or is this about the non-test service code?

service/net/tailscale.go
skhaz: Tor is a thin SOCKS5 wrapper (the only extra is stream isolation via random credentials), so it could be replaced by a generic SOCKS5 service with an optional isolation flag. However, the others are not simple proxies:

- **I2P** implements the SAM v3 protocol directly (two-connection handshake, session management, stream connect) — not SOCKS at all.
- **OpenVPN** queries the management interface to discover the VPN-assigned IP, then binds outgoing connections to that IP — also not SOCKS.

So a reasonable simplification would be: generic SOCKS5 service (covering Tor and any other SOCKS proxy) + I2P + OpenVPN + Tailscale. Want me to refactor Tor into a generic SOCKS5 service?

api/runtime/lua/config.go
skhaz: Done. Removed the top-level `Network` field from `FunctionConfig`, `ProcessConfig`, `BytecodeFunctionConfig`, and `BytecodeProcessConfig`. The network is now read from `Meta` via `cfg.Meta.GetString("network", "")`, consistent with how other options are handled.

skhaz: https://github.com/wippyai/runtime/pull/238/changes/17390ac245727eb021ec45ade1e63065ac9c3d0c

service/net/tailscale.go
wolfy-j: yes, no need to thin wrappers if lower level abstraction will work

api/runtime/lua/config.go
wolfy-j: I think we had delegated meta.options, but i will double check on merge.

PR #239, opened 2026-04-02
Add AMQP, SQS, and Redis Streams queue drivers

skhaz:
## Summary

Implement four new pluggable queue drivers following the existing memory driver pattern:

- **AMQP (RabbitMQ)**: uses `amqp091-go` for publish/consume via channels
- **AWS SQS**: uses `aws-sdk-go-v2` with long polling and message attributes
- **Redis Streams**: uses `go-redis/v9` with XREADGROUP consumer groups

Each driver includes config, driver, manager, util, and boot component.

## Driver Details

| Driver | Kind | Publish | Consume | Config |
|--------|------|---------|---------|--------|
| AMQP | `queue.driver.amqp` | `channel.Publish` | `channel.Consume` (push) | `url` |
| SQS | `queue.driver.sqs` | `SendMessage` | `ReceiveMessage` (long poll) | `region`, `endpoint` |
| Redis | `queue.driver.redis` | `XADD` | `XREADGROUP` (consumer groups) | `addr`, `password`, `db` |

## Example: AWS SQS

```yaml
entries:
  - name: sqs_driver
    kind: queue.driver.sqs
    region: us-east-1
    # endpoint: http://localhost:4566  # LocalStack
    lifecycle:
      auto_start: true

  - name: orders
    kind: queue.queue
    driver: app.queue:sqs_driver
    options:
      queue_name: my-orders-queue

  - name: order_handler
    kind: function.lua
    source: file://order_handler.lua
    modules: [queue, logger]

  - name: order_consumer
    kind: queue.consumer
    queue: app.queue:orders
    func: app.queue:order_handler
    concurrency: 5
    prefetch: 10
    lifecycle:
      auto_start: true
```

## Tests

Screenshot 2026-04-03 at 13 59 56


api/service/queue/sqs/config.go
```diff
@@ -0,0 +1,27 @@
...
+	"github.com/wippyai/runtime/api/registry"
+	"github.com/wippyai/runtime/api/supervisor"
+)
+
+// Kind identifies the AWS SQS queue driver.
+const Kind registry.Kind = "queue.driver.sqs"
+
+// Config defines the AWS SQS queue driver configuration.
+type Config struct {
+	Region    string                     `json:"region"`
```
skhaz: Just for clarification, is this AWS account ours, or can the client provide their own credentials?

api/service/queue/amqp/config.go
```diff
@@ -0,0 +1,34 @@
...
+	"github.com/wippyai/runtime/api/registry"
+	"github.com/wippyai/runtime/api/supervisor"
+)
+
+// Kind identifies the AMQP (RabbitMQ) queue driver.
+const Kind registry.Kind = "queue.driver.amqp"
+
+// Config defines the AMQP queue driver configuration.
+type Config struct {
+	URL       string                     `json:"url"`
```
wolfy-j: Can you check RR driver options as well, it should have sopme QoS options and possibly TSL.

api/service/queue/redis/config.go
```diff
@@ -0,0 +1,31 @@
...
+	"github.com/wippyai/runtime/api/registry"
+	"github.com/wippyai/runtime/api/supervisor"
+)
+
+// Kind identifies the Redis Streams queue driver.
+const Kind registry.Kind = "queue.driver.redis"
+
+// Config defines the Redis Streams queue driver configuration.
+type Config struct {
+	Addr      string                     `json:"addr"`
```
wolfy-j: Need all options support like Cluster and etc.

api/service/queue/sqs/config.go
wolfy-j: See AWS service, we already have centralized canonical pass to provide Cloud creds, you can use it to get clietn, so here you really just specify ID of aws config entry. And yes, anyone can provide own credentials.

service/queue/amqp/driver.go
```diff
@@ -0,0 +1,336 @@
...
+}
+
+func (d *Driver) channelFromConn(conn *amqp091.Connection) (*amqp091.Channel, error) {
+	if conn == nil || conn.IsClosed() {
+		return nil, queuesvc.ErrDriverNotStarted
+	}
+
+	ch, err := conn.Channel()
+	if err != nil {
+		return nil, fmt.Errorf("amqp channel: %w", err)
```
wolfy-j: api/error

service/queue/amqp/driver.go
```diff
@@ -0,0 +1,336 @@
...
+		if name := opts.GetString(queueapi.OptionQueueName, ""); name != "" {
+			return name
+		}
+	}
+	return queueID.Name
+}
+
+func (d *Driver) Publish(ctx context.Context, queueID registry.ID, msgs ...*queueapi.Message) error {
+	d.mu.RLock()
+	q, exists := d.queues[queueID]
```
wolfy-j, line 290: Gonna be blocking.

service/queue/amqp/driver.go
```diff
@@ -0,0 +1,336 @@
...
+	}
+
+	ch, err := d.getChannel()
+	if err != nil {
+		return nil, err
+	}
+
+	consumerTag := fmt.Sprintf("%s-%s", queueID.String(), uuid.New().String()[:8])
+	amqpDeliveries, err := ch.Consume(
+		q.name,      // queue
```
wolfy-j: can we configure all that via mapping from queue options? It already have property for driver level mapping.

service/queue/amqp/driver.go
```diff
@@ -0,0 +1,336 @@
...
+
+	qi, err := ch.QueueDeclarePassive(q.name, false, false, false, false, nil)
+	if err != nil {
+		return nil, fmt.Errorf("amqp queue inspect: %w", err)
+	}
+
+	info := attrs.NewBag()
+	info.Set(queueapi.StatsMessageCount, qi.Messages)
+	info.Set(queueapi.StatsConsumerCount, qi.Consumers)
+	info.Set(queueapi.StatsReady, qi.Messages)
```
wolfy-j, line 865: nice

service/aws/sqs/driver.go
```diff
@@ -0,0 +1,327 @@
+// SPDX-License-Identifier: MPL-2.0
+
+package sqs
```
wolfy-j, line 3: Must be withjing service aws, yes it's queue, but provider specific.

service/aws/sqs/util.go
```diff
@@ -0,0 +1,29 @@
...
+package sqs
+
+import (
+	"encoding/json"
+
+	"github.com/wippyai/runtime/api/payload"
+)
+
+// marshalBody converts a payload to JSON bytes for SQS message body.
+func marshalBody(p payload.Payload) ([]byte, error) {
```
wolfy-j: Check payload and trascoder please, we have canonical way to marshal payloads.

service/aws/sqs/util.go
```diff
@@ -0,0 +1,29 @@
...
+	if p == nil {
+		return []byte("null"), nil
+	}
+	return json.Marshal(p)
+}
+
+// unmarshalBody converts SQS message body bytes to a payload.
+func unmarshalBody(data []byte) payload.Payload {
```
wolfy-j: we have canonical way but for inteoperability it _does_ make sense to include per-queue option tospecific which codec it prefers.

Please check RR queues, they have a lot of years of hardening and option handling that otherwise we will have to re-discover.

wolfy-j:
We ideally need a harness for it as well.

api/service/queue/redis/config.go
skhaz: [Add full Redis connection options: standalone, cluster, and sentinel …](https://github.com/wippyai/runtime/pull/239/commits/0ca4840ac584e3dae16642cac71b5e00ecfd8aa6)

api/service/queue/amqp/config.go
skhaz: [Expand AMQP config with full connection options and TTL support, fix …](https://github.com/wippyai/runtime/pull/239/commits/ce2805d65b4d61dd4ee1385d1c42663bc43fd80c)

api/service/queue/sqs/config.go
```diff
@@ -0,0 +1,467 @@
...
+
+	// TLS configures TLS settings for the HTTP transport.
+	// When TLS.Enabled is true, a custom TLS configuration is applied.
+	TLS *TLSConfig `json:"tls,omitempty"`
+
+	// AWSConfig is an optional reference to a shared config.aws resource ID.
+	// When set, credentials and region are acquired from the shared resource
+	// (same pattern as the S3 driver). Inline Region/AccessKeyID fields are
+	// ignored when this is set.
+	AWSConfig string `json:"aws_config,omitempty"`
```
wolfy-j: registry.ID

service/queue/sqs/manager.go
```diff
@@ -0,0 +1,267 @@
...
+	if err != nil {
+		return aws.Config{}, fmt.Errorf("sqs: acquire aws config resource %q: %w", cfg.AWSConfig, err)
+	}
+
+	gotConfig, err := rsc.Get()
+	if err != nil {
+		return aws.Config{}, fmt.Errorf("sqs: get aws config resource: %w", err)
+	}
+
+	awsCfg, ok := gotConfig.(aws.Config)
```
wolfy-j: check existing client abstraction, we must uyse it here, similar to S3

service/queue/sqs/manager.go
```diff
@@ -0,0 +1,267 @@
...
+
+	if cfg.Region != "" {
+		opts = append(opts, awsconfig.WithRegion(cfg.Region))
+	}
+	if cfg.Profile != "" {
+		opts = append(opts, awsconfig.WithSharedConfigProfile(cfg.Profile))
+	}
+
+	// Static credentials
+	if cfg.AccessKeyID != "" && cfg.SecretAccessKey != "" {
```
wolfy-j: same, it's not place for it here

api/service/aws/sqs/config.go
```diff
@@ -0,0 +1,100 @@
...
+	MaxNumberOfMessages int32 `json:"max_number_of_messages,omitempty"`
+
+	// WaitTimeSeconds is the long-poll wait time in seconds (0–20).
+	// 0 means short polling. Default: 20.
+	WaitTimeSeconds int32 `json:"wait_time_seconds,omitempty"`
+
+	// VisibilityTimeout is the default visibility timeout in seconds (0–43200)
+	// applied per ReceiveMessage call.
+	// 0 uses the queue's default (typically 30s).
+	VisibilityTimeout int32 `json:"visibility_timeout,omitempty"`
```
wolfy-j: This is not per queue?

system/clock/timer.go
```diff
@@ -71,6 +71,13 @@ func (r *timerRegistry) startWithCallback(d time.Duration, callback func()) uint
 			return
 		}
 
+		// Entries without callbacks are removed by wait/stop to preserve explicit lifecycle.
+		// Delete before callback so that any observer reacting to the callback
+		// sees the entry already gone (e.g. stop() returns ErrTimerNotFound).
+		if entry.callback != nil {
```
wolfy-j, line 77: ??

system/clock/timer.go
```diff
@@ -71,6 +71,13 @@ func (r *timerRegistry) startWithCallback(d time.Duration, callback func()) uint
...
+			r.deleteEntry(shard, id)
+		}
+
 		if entry.callback != nil {
```
wolfy-j: not sure why it's in this PR, could be right but we dont need two ifs (thought compiler most likely will merge brances it reads weird

service/queue/redis/util.go
```diff
@@ -0,0 +1,77 @@
...
+	"strings"
+
+	goredis "github.com/redis/go-redis/v9"
+	"github.com/wippyai/runtime/api/attrs"
+	"github.com/wippyai/runtime/api/payload"
+	queueapi "github.com/wippyai/runtime/api/queue"
+)
+
+// marshalBody converts a payload to JSON bytes.
+func marshalBody(p payload.Payload) ([]byte, error) {
```
wolfy-j: redis only json? we can't use other configurable formats? transcoder support msgpack for example

service/queue/memory/driver_test.go
```diff
@@ -37,149 +46,6 @@ func TestMemoryDriver_DeclareQueue(t *testing.T) {
 	assert.Equal(t, 50, cap(q.messages))
 }
 
-func TestMemoryDriver_PublishAndAttach(t *testing.T) {
```
wolfy-j, line 40: any reason why this tests were removed?

service/queue/amqp/driver.go
skhaz: All `fmt.Errorf` returns in the AMQP driver have been replaced with `apierror.New()` using structured error kinds:

- **Unavailable** (with `WithRetryable(True)`) for connection, channel, publish, consume, declare, and inspect errors
- **Internal** for configuration/TLS build and marshal errors

This aligns with the canonical pattern established in `service/queue/errors.go`.

service/queue/amqp/driver.go
skhaz: `Publish()`, `Attach()`, and `GetQueueInfo()` now use `TryRLock()` instead of `RLock()`. When the write lock is held (e.g., during Stop/Start reconnect), callers receive a retryable `Unavailable` error instead of blocking indefinitely. This lets upstream code retry or back off gracefully.

service/queue/amqp/driver.go
skhaz: AMQP consume parameters are now configurable per-queue via queue options:

- `exclusive` — exclusive consumer
- `auto_ack` — automatic acknowledgment
- `no_local` — do not receive own publishes
- `no_wait` — do not wait for server confirmation
- `consumer_tag` — custom consumer tag prefix

All are read from `declaredQueue.opts` with sensible defaults (all false, auto-generated tag). Follows RoadRunner's per-pipeline configuration approach where receive-side parameters live alongside the queue declaration.

service/aws/sqs/util.go
skhaz: Replaced raw `encoding/json` with the canonical payload transcoder. `marshalBody` and `unmarshalBody` now accept a `payload.Transcoder` and a codec format string, transcoding through the graph-based routing system.

service/aws/sqs/util.go
skhaz: Added `OptionCodec` as a per-queue option in `api/queue/driver.go`. Each queue can now specify its preferred serialization format (e.g., `"json/plain"`, `"application/msgpack"`) — the transcoder handles multi-hop conversion via its graph-based routing. Default remains `json/plain` for backward compatibility.

Reviewed RoadRunner queue implementations — they use raw body passthrough with metadata in broker headers. Our approach is more flexible: the transcoder already supports JSON, MsgPack, YAML, and any future registered codec, with automatic format conversion between them.

api/service/queue/sqs/config.go
skhaz: Changed `AWSConfig` from `string` to `registry.ID`. This gives us type safety at the config level and proper JSON marshaling/unmarshaling via `registry.ID`'s built-in support (handles both `"ns:name"` string and `{"ns":"...","name":"..."}` object formats). The `registry.ParseID()` call in the manager is no longer needed — we use `cfg.AWSConfig` directly when acquiring the resource.

api/service/aws/sqs/config.go
skhaz: Moved `MaxNumberOfMessages`, `WaitTimeSeconds`, and `VisibilityTimeout` out of the driver-level `Config` and into per-queue options: `OptionMaxMessages`, `OptionWaitTime`, `OptionVisibilityTimeout` (defined in `api/queue/driver.go`). Each queue can now override these values via `DeclareQueue` opts, with sensible defaults (10, 20, 0 respectively). The `buildQueueAttributes()` method also reads `OptionVisibilityTimeout` from per-queue opts for the CreateQueue attribute.

system/clock/timer.go
skhaz: Good catch — this was an artifact of a race fix that moved `deleteEntry()` before the callback execution to prevent flaky `stop()` behavior, but accidentally left the original `if entry.callback != nil` as a separate block. Merged into a single block now: delete entry first, then execute callback.

system/clock/timer.go
skhaz: Fixed — the two `if` blocks are now merged into one. The timer fix is in this PR because the queue conformance test suite exercises timers for delayed delivery tests, and the flaky race surfaced during that work. But I can split it into a separate PR if you prefer.

service/queue/redis/util.go
skhaz: Integrated the payload transcoder into the Redis driver, same pattern as SQS. `marshalBody` and `unmarshalBody` now go through `tc.Transcode()` with the codec format from `OptionCodec` (defaults to JSON). This means Redis queues can use msgpack or any other registered format by setting `codec: "application/msgpack"` in the queue declaration options. Updated the driver, manager (passes transcoder), and tests.

api/service/queue/sqs/config.go
skhaz: Already using the centralized AWS config pattern — the SQS driver references a shared `config.aws` resource via the `AWSConfig` field (now `registry.ID` for type safety). At runtime, the manager acquires the AWS config through `resource.GetRegistry(ctx).Acquire()`, same pattern as the S3 driver. No inline AWS credentials in the SQS config.

service/aws/sqs/driver.go
skhaz: Resolved — the SQS driver implementation now lives in `service/aws/sqs/` under the AWS service namespace (alongside `service/aws/config/` and `service/aws/s3/`), while the config/API types are in `api/service/aws/sqs/`. This keeps it consistent with other AWS services in the codebase.

service/queue/sqs/manager.go
skhaz: The SQS manager uses the existing shared AWS config abstraction from `resource.GetRegistry(ctx)` — same `Acquire` pattern as the S3 manager. The AWS SDK client (`sqs.NewFromConfig`) is created from the acquired `aws.Config` with SQS-specific endpoint/option overrides applied on top. No custom client abstraction needed since the AWS SDK already provides the configurable interface.

service/queue/sqs/manager.go
skhaz: Resolved — removed all inline AWS credential handling (AccessKeyID, SecretAccessKey, Region, SessionToken, Profile). The SQS config now only has `AWSConfig registry.ID` which references a shared `config.aws` resource. The manager acquires the full `aws.Config` from the resource registry at runtime, same as S3.

service/queue/memory/driver_test.go
skhaz: The per-driver tests were refactored into a shared conformance test suite at `service/queue/drivertest/suite.go` with 28 test cases covering: publish/consume, multi-message batching, multi-consumer distribution, message ordering, ack/nack semantics, queue isolation, concurrent operations, empty queue behavior, unknown queue errors, header round-trips, DeclareQueue idempotency, GetQueueInfo accuracy, consumer cancel, large payloads, special characters, nil body, and more. Each driver (memory, AMQP, Redis, SQS) now runs the full suite via `drivertest.New(t, driver, ...opts).Run()` with driver-specific capability flags (e.g. `WithNackRedelivers(false)` for Redis). This ensures consistent behavior guarantees across all backends while eliminating test duplication.

api/queue/driver.go
```diff
@@ -15,12 +15,27 @@ const (
...
+
+	// Consume/receive options (per-queue, read by Attach)
+	OptionExclusive   = "exclusive"    // Exclusive consumer (AMQP)
+	OptionAutoAck     = "auto_ack"     // Automatic acknowledgment
+	OptionNoLocal     = "no_local"     // Do not receive own publishes (AMQP)
+	OptionNoWait      = "no_wait"      // Do not wait for server confirmation (AMQP)
+	OptionConsumerTag = "consumer_tag" // Custom consumer tag prefix (AMQP)
+
+	// Codec option (per-queue serialization format)
+	OptionCodec = "codec" // Payload format for message body (e.g. "json/plain", "application/msgpack")
```
wolfy-j: nice

wolfy-j (commented):
## Review: Queue Drivers

Good work on the shared conformance suite — 28 test cases across all drivers with capability flags (`preservesHeaders`, `nackRedelivers`, `supportsReattach`, etc.) is the right approach. The Docker-based test setup with auto-cleanup is solid.

### Config bloat

The Redis config is 384 lines and the AMQP config is 274 lines. Most of this is mapping every possible client option into a Wippy config struct with JSON tags, validation, and `InitDefaults()`. For example, the Redis config exposes `MasterName`, `SentinelPassword`, `RouteByLatency`, `RouteRandomly`, `MaxActiveConns`, `MinRetryBackoff`, `MaxRetryBackoff`, `DialTimeout`, `ReadTimeout`, `WriteTimeout`, `ContextTimeoutEnabled`, `PoolFIFO`, `PoolSize`, `MinIdleConns`, `MaxIdleConns`, `MaxActiveConns`, `ConnMaxIdleTime`, `ConnMaxLifetime`, and TLS config.

This is wrapping the `go-redis/v9` `UniversalOptions` struct field by field. If a user needs `RouteByLatency`, they can set it. But do we need to define, validate, document, and test each field? Consider accepting the upstream options struct directly (or a JSON-passthrough `map[string]any` for advanced options) and only pulling out the fields that Wippy needs to validate or transform. The SQS config is cleaner — only `Region`, `Endpoint`, `Lifecycle`.

### No connection recovery

The AMQP driver creates a single connection in `Start()` and uses it until `Stop()`. If RabbitMQ restarts or the network blips, the connection dies and all subsequent operations fail with "amqp driver busy (reconnecting)" — but there's no actual reconnection logic. The error message references reconnecting but the code doesn't reconnect.

AMQP connections drop routinely in production (server restarts, network partitions, policy changes). The driver should either:
1. Reconnect automatically with backoff (watch `conn.NotifyClose()`, redial, re-declare queues)
2. Or at minimum, document that connection recovery is the supervisor's responsibility (stop + restart the service)

The Redis driver uses `go-redis` which has built-in reconnection. SQS is HTTP-based so no persistent connection. AMQP is the only one that needs explicit handling.

### Redis Streams: no pending message recovery

When a consumer crashes after receiving a message but before ACKing, the message stays in the Redis Stream's Pending Entries List (PEL). The current driver only reads new messages (`>`). It never claims pending messages (`XAUTOCLAIM` or `XCLAIM`). After a consumer restart, those messages are stuck forever.

Production Redis Streams consumers need a recovery loop:
1. On attach, first claim pending messages: `XAUTOCLAIM stream group consumer min-idle-time 0-0`
2. Process those before switching to `>` for new messages
3. Periodically sweep pending entries for messages abandoned by dead consumers

The Nack implementation also just logs — it doesn't requeue or mark for retry. The comment says "No XACK means the message will be redelivered via pending entries" but nothing actually reclaims those pending entries.

### AMQP: channel-per-operation is expensive

`getChannel()` / `getChannelLocked()` creates a new AMQP channel for each `Publish`, `DeclareQueue`, and `GetQueueInfo` call, then immediately closes it:

```go
ch, err := d.getChannel()
// ... use channel ...
defer ch.Close()
```

AMQP channels are multiplexed over the connection but still have setup/teardown cost (server-side resource allocation, handshake). For high-throughput publishing, this is a bottleneck. Consider:
- A persistent publish channel (with reconnect on channel error)
- Or a channel pool (bounded, reusable)

The `Attach` consumer correctly keeps its channel open for the lifetime of the consumer — apply the same pattern to publish.

### SQS: batch publish not used

SQS supports `SendMessageBatch` for up to 10 messages per API call. The current implementation sends one `SendMessage` per message in a loop. For the `Publish(ctx, queue, msgs...)` variadic API, batch publishing would reduce API calls significantly.

### Queue option constants growing unbounded

`api/queue/driver.go` now has ~30 option constants mixing generic options (`OptionQueueName`, `OptionDurable`), AMQP-specific options (`OptionExclusive`, `OptionNoLocal`, `OptionConsumerTag`), and SQS-specific options (`OptionMaxMessages`, `OptionVisibilityTimeout`). Plus reserved-for-future comments.

Driver-specific options should live in their respective config packages, not in the shared `api/queue/` package. The shared package should only define options that apply to ALL drivers. Otherwise every new driver adds its options to the shared package and the shared API becomes a union of all driver quirks.

### Test harness: Docker management in test code

`drivertest/docker.go` has `DockerAvailable()` and `WaitForPort()`. Each driver's `TestMain` does its own `docker run` / `docker rm`. This works but consider:
- Using `testcontainers-go` which handles lifecycle, port mapping, and cleanup automatically
- Or at least extracting the container lifecycle into the harness so each driver test doesn't duplicate the docker run/rm pattern

### Minor

- `system/clock/timer.go` change (-5 +4 lines) is unrelated to queue drivers — should be a separate commit or PR.
- The `drivertest.Harness` is well-designed. The capability flags approach (each driver declares what it supports) is clean and extensible.

api/service/aws/sqs/config.go
```diff
@@ -0,0 +1,72 @@
...
+// and other AWS SDK settings.
+type Config struct {
+	// AWSConfig is a resource ID referencing a config.aws resource
+	// managed by service/aws/config/manager.go.
+	// Resolved at runtime via resource.GetRegistry(ctx).Acquire().
+	AWSConfig registry.ID `json:"config"`
+
+	// Endpoint is a custom endpoint URL (e.g. for LocalStack, ElasticMQ).
+	// Sets BaseEndpoint on the AWS config.
+	Endpoint string `json:"endpoint,omitempty"`
```
wolfy-j, line 30: any reason it's there since you have client?

api/service/queue/redis/config.go
```diff
@@ -0,0 +1,384 @@
...
+
+	// SentinelUsername for ACL-based authentication with sentinel nodes.
+	SentinelUsername string `json:"sentinel_username,omitempty"`
+
+	// SentinelPassword for authentication with sentinel nodes.
+	SentinelPassword string `json:"sentinel_password,omitempty"`
+
+	// MinRetryBackoff is the minimum backoff between retries.
+	// -1 disables backoff. Default: 8ms.
+	MinRetryBackoff time.Duration `json:"min_retry_backoff,omitzero,format:units"`
```
wolfy-j: please nest retry policy, also what this retry policy is for? redis or queue?

service/queue/drivertest/harness.go
```diff
@@ -0,0 +1,140 @@
+// SPDX-License-Identifier: MPL-2.0
```
wolfy-j, line 1: build flags?

service/queue/drivertest/docker.go
```diff
@@ -0,0 +1,49 @@
+// SPDX-License-Identifier: MPL-2.0
```
wolfy-j, line 1: build flags?

service/queue/drivertest/suite.go
```diff
@@ -0,0 +1,1060 @@
+// SPDX-License-Identifier: MPL-2.0
```
wolfy-j, line 1: biuld flags?

service/queue/redis/driver_test.go
```diff
@@ -0,0 +1,93 @@
...
+	"github.com/stretchr/testify/require"
+	"github.com/wippyai/runtime/api/registry"
+	"github.com/wippyai/runtime/service/queue/drivertest"
+	systempayload "github.com/wippyai/runtime/system/payload"
+	sysjson "github.com/wippyai/runtime/system/payload/json"
+	"go.uber.org/zap/zaptest"
+)
+
+const (
+	testContainer = "wippy-test-redis"
```
wolfy-j: do we have it docker template we had in tests?

service/queue/redis/util.go
```diff
@@ -0,0 +1,119 @@
...
+func marshalBody(tc payload.Transcoder, codec string, p payload.Payload) ([]byte, error) {
+	if p == nil {
+		return []byte("null"), nil
+	}
+
+	// Ensure we start from a golang/any payload
+	src := p
+	if src.Format() != payload.Golang {
+		var err error
+		src, err = tc.Transcode(src, payload.Golang)
```
wolfy-j: what is the reason for it?

service/queue/redis/util.go
```diff
@@ -0,0 +1,119 @@
...
+		return nil, fmt.Errorf("transcode to %s: %w", codec, err)
+	}
+
+	switch v := encoded.Data().(type) {
+	case []byte:
+		return v, nil
+	case string:
+		return []byte(v), nil
+	default:
+		return nil, fmt.Errorf("transcoded payload data is not []byte or string, got %T", v)
```
wolfy-j: we have error class for this i guess

service/queue/redis/util.go
```diff
@@ -0,0 +1,119 @@
...
+func unmarshalBody(tc payload.Transcoder, codec string, data []byte) payload.Payload {
+	if len(data) == 0 {
+		return nil
+	}
+
+	// Create a payload in the source codec format
+	src := payload.NewPayload(data, codec)
+
+	// Transcode to golang/any
+	result, err := tc.Transcode(src, payload.Golang)
```
wolfy-j: why?

service/queue/redis/util.go
```diff
@@ -0,0 +1,119 @@
...
+	if id, ok := redisMsg.Values["id"].(string); ok {
+		msg.ID = id
+	}
+
+	if body, ok := redisMsg.Values["body"].(string); ok {
+		msg.Body = unmarshalBody(tc, codec, []byte(body))
+	}
+
+	if headers, ok := redisMsg.Values["headers"].(string); ok {
+		msg.Headers = unmarshalHeaders([]byte(headers))
```
wolfy-j: is it normal convention in redis?

wolfy-j (changes requested):

service/queue/amqp/driver.go
wolfy-j: }
<-- what if lock appears here?
	q, exists := d.queues[queueID]

service/queue/amqp/driver.go
```diff
@@ -0,0 +1,489 @@
...
+	return neverClosedChan
+}
+
+func (d *Driver) Start(ctx context.Context) (<-chan any, error) {
+	amqpCfg, err := d.buildAMQPConfig()
+	if err != nil {
+		return nil, apierror.New(apierror.Internal, "amqp config").WithCause(err)
+	}
+
+	conn, err := amqp091.DialConfig(d.cfg.URL, amqpCfg)
```
wolfy-j, line 889: reconnects?

service/aws/sqs/driver.go
skhaz: Already done — SQS driver was moved from `service/queue/sqs/` to `service/aws/sqs/` in the initial reorganization commit (dc0c7f80). Package path is now `service/aws/sqs/`, matching the provider-specific pattern.

service/queue/sqs/manager.go
skhaz: Done — the SQS manager now uses the shared AWS config resource abstraction via `resource.GetRegistry(ctx).Acquire()`, same pattern as the S3 manager. The `AWSConfig` field is `registry.ID` pointing to the centralized AWS config entry.

service/queue/sqs/manager.go
skhaz: Agreed and done — SQS manager is now in `service/aws/sqs/manager.go` and uses the shared AWS config resource pattern, no client construction inside the manager.

service/queue/redis/util.go
skhaz: Fixed — Redis (and SQS) now use the canonical payload transcoder with per-queue codec support. Each queue can specify its preferred format via the `OptionCodec` option (e.g. `"json/plain"`, `"application/msgpack"`). The transcoder handles graph-based conversion between any registered formats. Default remains `json/plain` for backward compatibility.

See `service/queue/redis/util.go` — `marshalBody()` and `unmarshalBody()` now go through `tc.Transcode()` with the queue's codec, and `queueCodec()` reads `OptionCodec` from queue opts.

service/queue/amqp/driver.go
skhaz: Commit: fa2d4495

service/queue/amqp/driver.go
skhaz: Commit: 193849ab

service/queue/amqp/driver.go
skhaz: Commit: aa17dea7

system/clock/timer.go
skhaz: Commit: 7c3f3c4b

PR #241, opened 2026-04-08
Cluster primitives

skhaz:
Implement distributed named process groups following Erlang/OTP pg semantics.

## Matching Semantics with Erlang pg

| Feature | Erlang pg | This implementation |
|---|---|---|
| Named groups | Processes join arbitrary named groups | Same (Group = string) |
| Multi-join | A process can join the same group multiple times and must leave the same number of times | Same (state tracks duplicate group entries) |
| Auto-removal on exit | Processes are automatically removed from all groups when they terminate | Same (via monitorProcess / handleProcessExit) |
| Cluster-wide membership | Group membership is visible across all connected nodes | Same (local + remote state, discover/sync protocol) |
| Eventually consistent | No strong consistency guarantees; nodes converge via async messaging | Same (async broadcast of join/leave/sync messages) |
| get_members / get_local_members | Separate queries for all-nodes vs local-node members | Same (GetMembers vs GetLocalMembers) |
| which_groups | Returns all groups with at least one member | Same |

## Differences from Erlang pg

| Feature | Erlang pg | This implementation |
|---|---|---|
| Scopes | Supports named scopes (multiple independent pg instances) | Single global scope, no scope parameter |
| Monitor callbacks | pg:monitor/1,2 lets a process receive join/leave notifications | No equivalent, query-only |
| Broadcast | No built-in broadcast; sending to members is left to the caller | Adds Broadcast and BroadcastLocal as first-class operations |
| Consistency model | gen_server with ordered message processing per-scope | Single-goroutine event loop via actions channel (functionally equivalent) |


wolfy-j (commented):
## Review: Architecture Alignment

The core PG implementation is well-built. Good.

### PG should follow process host semantics

Currently PG is a hardcoded singleton created directly in the boot component. Every other service that runs processes follows the **registry entry → manager → named instances** pattern:

- `service/host/` — process hosts are declared as registry entries with `kind: "process.host"`, managed by `host.Manager` which implements `registry.EntryListener` (Add/Update/Delete). Each entry creates a `Host` with its own scheduler, its own relay host registration (`HostID = entry.ID`), its own supervisor lifecycle.
- `service/store/memory/` — memory stores follow the same pattern with `kind: "store.memory"`.
- `service/sql/` — SQL connections follow the same pattern.

PG should do the same:

```yaml
# Registry entry declares a named PG instance
- id: "pg:chat"
  kind: "pg.host"
  config:
    queue_size: 1024 # and other valuable options if any

- id: "pg:presence"
  kind: "pg.host"
```

A `PGManager` handles `Add/Update/Delete`:
- Creates a PG `Service` per entry
- Registers as relay host with `HostID = entry.ID.String()` (e.g., `"pg:chat"`)
- Registers with supervisor for lifecycle management
- Registers as **resource provider** so Lua code can acquire it

This gives you:
1. **Multiple named instances** with real isolation (separate state, separate event loops)
2. **Security boundaries** — Lua code acquires a named PG via resource registry with `security.IsAllowed(ctx, "pg.open", "pg:chat")`, same pattern as `store.get` and `db.get`
3. **Proper lifecycle** — supervisor manages start/stop/restart per instance
4. **No singleton context hack** — `pgapi.GetProcessGroups(ctx)` goes away, replaced by resource acquisition

Reference implementation: `service/host/manager.go` lines 53-101 (Add method) shows the exact pattern — decode config, create instance, fire `relay.HostRegister` event, fire `supervisor.ServiceRegister` event.

### Replace synthetic PIDs with service addressing

`protocol.go` creates fake PIDs to route inter-node messages:

```go
func pgPID(nodeID pid.NodeID) pid.PID {
    return pid.PID{Node: nodeID, Host: pgapi.HostID, UniqID: "pg"}
}
```

These aren't real processes — they're service addresses. They pollute the PID space and could accidentally interact with topology (monitors, links). The relay system (`system/relay/node.go` line 75) already routes purely by `pkg.Target.Host` — the `UniqID` is irrelevant for host-level routing.

Add a `relay.NewServicePackage(sourceNode, sourceHost, targetNode, targetHost, topic, payloads...)` constructor that creates packages with empty `UniqID`. This makes the intent clear and avoids synthetic PIDs. The internode codec already handles empty UniqID correctly.

With host-based PG, each instance gets its own HostID (e.g., `"pg:chat"`), so inter-node routing naturally targets the correct instance on the remote node.

### Lua surface: resource acquisition instead of singleton + prefix scoping

Current approach:
```lua
local pg = require("pg")
local s = pg.scope("chat")   -- prefix hack, same underlying service
s.join("room:123")            -- actually joins "chat::room:123" 
```

Target approach (matches store/db patterns):
```lua
local pg = require("pg")
local chat = pg.open("pg:chat")     -- acquires named instance via resource registry
chat:join("room:123")                -- real isolation, security check on "pg:chat"
```

Reference: `runtime/lua/modules/store/store.go` lines 100-141 shows the exact pattern — `security.IsAllowed(ctx, "store.get", id)` → `resource.GetRegistry(ctx).Acquire(resID)` → wrap in userdata.

The current `pg.scope()` prefix approach works functionally but doesn't provide real security boundaries — a process that knows another scope's prefix can reach across. Named instances with resource-level access control fix this.

### Extract replication transport (future-facing)

The discover/sync/delta-broadcast protocol in `protocol.go` + `service.go` is generic:
1. Subscribe to `cluster.NodeJoined` / `cluster.NodeLeft`
2. On join → discover → sync full state
3. On local mutation → broadcast delta to all known remotes
4. On node left → clean up remote state

This exact protocol is needed for distributed KV scopes (the next coordination primitive). Consider structuring the protocol code so it can be extracted into a shared `system/dds/` (distributed data sync) module later. This doesn't need to happen now, but keeping the protocol logic cleanly separated from PG's state management (which it mostly already is — `protocol.go` vs `state.go`) makes the future extraction straightforward.

### Config for pg.host

Minimal config per instance:

```go
type Config struct {
    Lifecycle          supervisor.LifecycleConfig `json:"lifecycle"`
    QueueSize          int                        `json:"queue_size"`          // action channel buffer (default 256)
    MonitorBuffer      int                        `json:"monitor_buffer"`      // monitor event channel buffer (default 64)
    MaxGroups          int                        `json:"max_groups"`          // 0 = unlimited
    MaxMembersPerGroup int                        `json:"max_members_per_group"` // 0 = unlimited
}
```

skhaz (commented):
Found a bug in `boot/components/core/infrastructure.go`: the `PIDGen()` component was creating the PID generator with an empty string as the node ID. Every PID in the system had a blank `Node` field, so remote nodes could never determine which node a PID belongs to. Fixed by reading the node ID from the relay node in context before constructing the generator.

wolfy-j:
Runtime integration review for cluster primitives:

The primitive family is the right direction, but I would tighten the public semantics around four areas before relying on it as the canonical distributed substrate.

1. Name-based process messaging

`process.send(destination, topic, ...)` should be the normal app-facing API and must resolve both layers internally:

- raw PID string: send directly
- global name: resolve through the global registry first
- local name: fallback to the local PID registry

If a global name is resolved, the runtime should attach and validate the ownership epoch/fence internally. App code should not normally call `lookup_with_fence` or `validate_fence`. Those can exist as low-level/debug APIs, but the typed public surface should make `send(name, ...)` safe by default. The process module type manifest should match whatever Lua API is exposed.

2. Global registration modes

I would split global registration into explicit scopes/modes:

- `process.registry.GLOBAL`: scalable global name. Raft quorum commit establishes the unique owner and current fence/epoch. This is linearizable ownership; stale sends are rejected by fence validation.
- `process.registry.ROOT` or `process.registry.GLOBAL_SYNCED`: stricter root/control-plane name. Same Raft ownership, but activation requires all live nodes in the membership snapshot to acknowledge the committed epoch within a timeout.
- `EVENTUAL` or equivalent: high-cardinality soft state, presence, sessions, caches, and user-level discovery. This should use CRDT/eventual semantics and must not be treated as singleton authority.

Suggested strict-mode state machine:

1. Commit `pending(name, pid, epoch, required_nodes, deadline)` to Raft. The name is reserved and locked, but not authoritative yet.
2. Nodes in the membership snapshot observe/apply the epoch and ack via the node mesh.
3. Acks are recorded through Raft.
4. Once `required_nodes` is covered, commit `active(name, epoch)`.
5. If the deadline expires first, commit `expired(name, epoch)` and release/fail the pending registration.

Future joiners are not part of the old ack set, but they must catch up to the current registry state before being marked ready for authority-bearing traffic. This keeps the expensive all-live-node propagation guarantee limited to small root/control scopes, while normal global names remain scalable.

3. Correctness from any node

Global registration must be correct when called from any node, not only from the current Raft leader. The current follower forwarding path appears to return only error/no-error and drops the typed FSM response. Callers such as `Register` expect `*RegisterResult`, so a follower-side successful registration can apply on the leader but still fail locally as an unexpected response type. The forwarding path should return typed command responses, or reconstruct the correct per-command response semantics.

4. Raft transport and topology

Raft should run over the Wippy node mesh / internode layer, using the same node-level membership/events as the rest of the runtime. It should not create a separate default control-plane TCP transport with separate failure semantics.

The desired abstraction is intra-runtime Raft over the native mesh:

- one cluster topology source of truth
- node events drive voter reconciliation and readiness
- global process registry and strong KV share the same mesh failure model and observability as process messaging and process groups
- node metadata may contain routing hints, but should not become a parallel raft-specific network model

5. KV split

The `api/kv` split is useful:

- `kvraft`: strongly-consistent KV for config/control data, CAS, and linearizable reads
- `kveventual`: CRDT/eventual KV for presence/cache/session-style state

This should remain a tiered abstraction so callers choose authority vs convergence explicitly.

Tests I would want before considering this canonical:

- concurrent global registration of the same name from multiple nodes
- global registration called from a follower node
- `process.send(global_name, ...)` resolves global first, local second
- stale fenced sends are rejected after re-registration
- strict/root registration reaches active only after all required live-node acks
- strict/root registration timeout expires without creating a second owner
- node partition/rejoin cleanup and catch-up before ready
- mesh-backed Raft transport under node leave/rejoin
- `kvraft` CAS/read-barrier behavior across nodes
- `kveventual` convergence and tombstone behavior under churn

skhaz:
## Bugs fixed

**1. Raft mesh transport nil-map panic**
`system/raft/mesh_transport.go:266`: `Close()` set `l.sessions = nil`, but late inbound `ClassRaftMesh` frames still arrived from the connection manager and tried to write to the zeroed map. Fixed with a `closeCh` early bail at the top of `getOrCreateSession` plus a re-check of `sessions == nil` under the write lock. Regression test `TestMeshTransport_InboundAfterCloseDoesNotPanic` added.

**2. `desiredVoterCount(2, _) = 1` demoted a healthy survivor**
`system/raft/selection.go`: the "round even down" rule collapsed the post-failure `{1,2}` quorum to `{1}`, triggering a leadership transfer + self-demote that took the cluster unreachable for seconds. Fixed: `eligible == 2 → 2`. This was the root cause of the `run_chaos.sh` hang.

**3. `TestSupervisor_ServiceTimeout` flake under `-race`**
8s ctx deadline left only ~3s of slack vs `StartTimeout=5s + startDelay=6s`. Loosened to 15s.

**4. Lua `transient_error()` missed retryable Raft errors**
`tests/proof/lua/{follower_register,lookup_options,root_lock_driver,partition_catchup}.lua` aborted on the first transient leader-change error. Added: `leadership lost`, `ErrLeadershipLost`, `not the raft leader`, `ErrNotLeader`, `leadership transfer in progress`, `transfer in progress`, `timed out`. Added a `register_with_retry` helper to the T4 driver harness.

**5. Legacy `api/raft.Config` fields had no deprecation markers**
`BindAddr` / `BindPort` / `AdvertiseAddr` / `AutoPort` remain in the struct during the deprecation cycle but the mesh transport ignores them. Added `// Deprecated:` doc comments + scoped `//nolint:staticcheck` justifications on the legacy TCP-fallback callers and the diagnostic log lines.

## Reviewer-asked tests now landed

| # | Test | Where |
|---|---|---|
| 1 | concurrent CONSISTENT cross-node | `system/globalreg/scope_table_test.go::TestConsistentScope_ConcurrentCrossNode` |
| 7 | partition / rejoin catch-up | `tests/proof/run_partition_catchup.sh` (untracked, like every chaos rig in this repo) |
| 8 | NodeLeft → RemoveServer | `system/raft/membership_test.go::TestRunReconcileOnce_NodeLeftRemovesServer` |
| 9 | kvraft CAS / read-barrier across nodes | `system/kvraft/cross_node_test.go::TestKVRaft_CASLinearizableAcrossNodes` |
| 10 | kveventual convergence + tombstone GC | `system/kveventual/cross_node_test.go::TestKVEventual_ConvergenceAcrossNodes` |

## Validation gate (local, HEAD = `4adec22a`)

- `go test -count=1 -race ./...` — 275 packages, 0 failures
- `golangci-lint run ./...` — 0 issues
- All 9 chaos rigs PASS in serial: `run_proof.sh`, `run_pg.sh`, `run_name_send.sh`, `run_root_scope_lock.sh`, `run_raft_over_mesh.sh`, `run_follower_register.sh`, `run_lookup_options.sh`, `run_chaos.sh`, `run_partition_catchup.sh`

PR #262, opened 2026-05-01
CLI: wippy publish --create + local-stack support

skhaz:
## Summary
- New `wippy publish --create` flag registers the module on the registry before initiating publish (one-shot publish for new modules)
- New `--module-visibility` (public/private, default private), `--module-type`, `--module-display-name` flags
- Auth + hub clients: allow `host.docker.internal` / `host.containers.internal` plaintext (for local Docker stacks where the CLI runs in a container)

## Test plan
- [x] `wippy auth login --token wpy_... --registry http://host.docker.internal:8082 --local` succeeds from inside Debian arm64 container
- [x] `wippy publish --create --module-visibility public` registers + publishes in one shot
- [x] Same flow with --module-visibility private

### DIFF
--- a/boot/deps/auth/client.go
+++ b/boot/deps/auth/client.go
@@ -35,8 +35,12 @@ func NewClient(baseURL string) (*Client, error) {
 		return nil, fmt.Errorf("invalid registry URL: %w", err)
 	}
 
-	// Require HTTPS for non-localhost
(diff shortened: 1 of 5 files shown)


wolfy-j:
I'm not sure i like the idea of programmatic module creation. It's not technical problem, need to talk about.

wolfy-j (approved):

wolfy-j (approved):

PR #267, opened 2026-05-12
fs/directory: resolve module-owned relative paths against module source root

skhaz:
Module-owned `fs.directory` entries with a relative `directory` path (e.g. `./public`) were resolving against the wippy process CWD instead of the module's source root, so modules loaded via `wippy.lock` `replacements:` failed to find their files.

Fix: when an entry carries a `module` meta attribute and `base` is not explicitly set to `project`, the resolver now joins the path with the module's registered `SourceRoot`. Vendor unpack writes the path in the same module-relative form (`directory: <name>`, `base: module`) so both code paths agree.

### DIFF
--- a/cmd/internal/entries/extract.go
+++ b/cmd/internal/entries/extract.go
@@ -17,9 +17,7 @@ import (
 
 // ExtractWappToDir extracts a .wapp file into a source directory with _index.yaml files
 // and source files. After extraction, the .wapp file is removed.
-// projectRoot is the directory containing wippy.lock, used to compute relative paths
-// for fs.directory entries reconstructed from embedded resources.
-func ExtractWappToDir(wappPath, targetDir, projectRoot string) error {
+func ExtractWappToDir(wappPath, targetDir string) error {
 	// Open and keep file open throughout extraction so resource FS handles remain valid
 	file, err := os.Open(wappPath)
 	if err != nil {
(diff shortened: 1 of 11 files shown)


cmd/internal/entries/extract.go
```diff
@@ -110,7 +110,7 @@ type extractedResource struct {
 // restoreEmbeddedResources converts fs.embed entries back to fs.directory by extracting
 // their matching resource filesystems to named subdirectories. Returns the modified entries
 // and any unclaimed resources.
-func restoreEmbeddedResources(entries []wapp.Entry, resources []extractedResource, targetDir, projectRoot string) ([]wapp.Entry, []extractedResource, error) {
+func restoreEmbeddedResources(entries []wapp.Entry, resources []extractedResource, targetDir, _ string) ([]wapp.Entry, []extractedResource, error) {
```
wolfy-j: maybe DTO at this level?

wolfy-j:
GG but see comment

cmd/internal/entries/extract.go
skhaz: Done in 29f80167. Introduced a private `restoreInput` DTO and used it to drop the dead `projectRoot` parameter from `ExtractWappToDir` and its callers — keeping the function inputs grouped without carrying around an unused field.

PR #269, opened 2026-05-14
client: route publish/install through hub-mediated endpoints with retry

skhaz:
## Problem

Windows users on flaky networks have been unable to publish for 10+ days. The legacy presigned-URL flow does `client → S3` directly for the wapp upload; a winsock reset (`wsarecv: Connection reset by peer`) aborts the upload with no retry. Same shape failure exists on the install/download side.

## Fix

Prefer the new hub-mediated endpoints:

- `POST /api/v1/publish/upload` — body is the wapp bytes; hub validates digest/size/auth, uploads to S3 with hub-side retry, starts the existing PublishWorkflow.
- `GET /api/v1/wapp/{digest}` — hub streams from S3 with retry, writes to client.

One HTTPS hop to the hub instead of two (RPC + presigned PUT). Falls back to the legacy `InitiatePublish → presigned PUT → ConfirmPublish` triplet if the hub doesn't advertise the new endpoints (404).

Retry policy mirrors the hub side — 8 attempts × jittered exponential backoff up to 30s, classifies retryable: winsock variants (`wsarecv:`, `wsasend:`), `connection reset`, `broken pipe`, EOF, timeouts, HTTP 429/5xx. `GetManifest` also wrapped.

## Test plan

- [x] `go test ./...` clean
- [x] `golangci-lint run ./boot/deps/hub/... ./cmd/...` clean
- [x] Cross-compiled `wippy.exe` (PE32+ x86_64, mingw + CGO sqlite_vec) under wine64/Rosetta in OrbStack: publish + install against local hub both hit the new endpoints (200 OK in hub access logs).

### DIFF
--- a/boot/deps/hub/errors.go
+++ b/boot/deps/hub/errors.go
@@ -17,7 +17,33 @@ var (
 	ErrDigestMismatch    = errors.New("digest mismatch")
 	ErrUploadExpired     = errors.New("upload URL expired")
 	ErrPublishInProgress = errors.New("publish already in progress")
+	ErrQuotaExceeded     = errors.New("quota exceeded")
 )
+
+type QuotaExceededError struct {
+	Reason string
+}
+
+func (e *QuotaExceededError) Error() string {
+	if e.Reason == "" {
+		return ErrQuotaExceeded.Error()
+	}
+
+	return ErrQuotaExceeded.Error() + ": " + e.Reason
+}
+
+func (e *QuotaExceededError) Is(target error) bool {
+	return target == ErrQuotaExceeded
+}
+
+func QuotaReason(err error) string {
+	var qe *QuotaExceededError
+	if errors.As(err, &qe) {
+		return qe.Reason
+	}
+
+	return ""
+}
 
 func MapConnectError(err error) error {
 	if err == nil {
@@ -38,6 +64,8 @@ func MapConnectError(err error) error {
 		return ErrModuleNotFound
 	case connect.CodeAlreadyExists:
 		return ErrVersionExists
+	case connect.CodeResourceExhausted:
+		return &QuotaExceededError{Reason: connectErr.Message()}
 	case connect.CodeInvalidArgument:
 		if containsMessage(connectErr, "version") {
 			return ErrInvalidVersion
--- a/boot/deps/hub/errors_test.go
+++ b/boot/deps/hub/errors_test.go
@@ -4,6 +4,7 @@ package hub
 
 import (
 	"errors"
+	"fmt"
 	"testing"
 
 	"connectrpc.com/connect"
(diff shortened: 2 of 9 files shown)


skhaz:
### Repro note: `GODEBUG=asyncpreemptoff=1` required on Apple Silicon

Verified the test plan end-to-end against `https://hub.wippy.ai` (prod-gen2 cluster) using the binary built at `a3110d4e`:

```
file dist/wippy-windows-amd64.exe
# PE32+ executable (console) x86-64 (stripped to external PDB), for MS Windows
sha256sum dist/wippy-windows-amd64.exe
# fb1ed53d8ebb15b2c073b1d0a812b773653902b32cc8ce149014b1a70f14c2ac
```

**Without** `GODEBUG=asyncpreemptoff=1` the binary aborts immediately on Rosetta:

```
0124:err:winediag:nodrv_CreateWindow ...
0124:err:ntlm:ntlm_LsaApInitializePackage no NTLM support, expect problems
rosetta error: invalid gdt selector index 5
```

Go's signal-based async preemption uses `fs`/`gs` segment selectors that Rosetta does not translate; the SIGURG handler trips the GDT check during scheduler init, before any user output. Disabling async preemption sidesteps the path.

**With** the env var set, both hub-mediated paths work end-to-end via `wine` (10.0) inside an OrbStack `wine` VM:

```sh
cp dist/wippy-windows-amd64.exe /Users/me/wippy.exe
export WIPPY_TOKEN=wpy_...
orb -m wine bash -c '
  GODEBUG=asyncpreemptoff=1 WINEDEBUG=-all \
    wine /Users/me/wippy.exe publish --create --module-visibility public \
      --registry https://hub.wippy.ai
'
```

Smoke result against prod hub:

| step | result |
|---|---|
| `wippy.exe version` | banner prints, Go runtime up |
| `publish --module-visibility public` | `loop-probe-2026/wine-smoke-public@0.0.1` registered + uploaded + processed (clean by ClamAV) |
| `publish --module-visibility private` | same for `loop-probe-2026/wine-smoke-private@0.0.1` |
| fresh dir `init` + `add` + `install --refresh` | both modules downloaded (private fetch proves token scope works) |
| `registry list` | both `ns.definition` entries present |

Hub workflow logs (`kubectl -n wippy-production logs -l app.kubernetes.io/name=hub`) show `publish workflow started` → `upload signal received` → `ValidateAndExtractWAPP` → `CreateVersion` → `ScanArtifact (clean)` → `UpdateModuleInfo` → `UpdateModuleStats` → `publish workflow completed` for both runs. No `s3.amazonaws.com` connect attempts visible — bytes flowed through `POST /api/v1/publish/upload`, not the legacy presigned PUT.

Suggest adding `GODEBUG=asyncpreemptoff=1` to the test-plan snippet in the PR body (or to a `docs/testing-windows.md`) so future contributors do not lose 30+ minutes chasing the GDT selector message.

PR #270, opened 2026-05-20
registry/supervisor: deterministic boot ordering

skhaz:
## Problem

`wippy run` boot intermittently failed with `failed to load state: operation failed for entry X: filesystem not found` (or `driver not found`), where X varied between runs of identical code. The failure was nondeterministic — same manifest, same binary, different outcome run-to-run.

## Root cause

Two cooperating issues:

1. **Nondeterministic boot ordering.** Five Go `range map` iterations leaked the process hash seed into the order the registry, supervisor, and topology builder scheduled work. With map iteration randomized, the order entries were dispatched to listeners changed every boot.

2. **Undeclared cross-entry dependencies.** Some manager listeners do synchronous lookups against sibling registry entries inside `Add` (e.g. `service/queue/consumer/manager.go:74-89` calls `m.queueMgr.GetQueue` / `m.queueMgr.GetDriver`). When the consumer entry was dispatched before its provider in a given run, the lookup missed and the listener rejected with `apierror.NotFound`, aborting the whole `Apply`.

The two issues compound: the random order means the rejection happens only sometimes; the missing declared dependency means the random order has nothing to constrain it.

## What this PR ships

**Layer 1 — deterministic ordering.** Sort the five randomized iteration sites by stable lexicographic keys:
- `supervisor.execute()` and `regTx.commit()` — sort `tx.register` / `tx.remove` / controllers maps
- `topology.Resolver.fetchDeps()` — sort output
- `topology.SquashChangesets()` and `topology.SortChangeSet()` — normalize input by `(NS, Name, Kind)` so stable-topological-sort tie-breaks don't leak the upstream slice's random order

Covered by N=1000 shuffle-property regression tests per fix site.

**Layer 2 — self-healing apply loop.** `BusRunner.Transition` now defers operations that reject with `apierror.NotFound` (the listener errors above) and retries them in a fixed-point loop after the rest of the pass completes. The loop terminates when either every op is accepted, an op fails with a non-deferrable error (rolls back as before), or a full pass makes zero progress (returns `NewUnresolvedDependenciesError`).

Sort alone was empirically insufficient: it just freezes the same flake into a deterministic always-fail when the consumer's name lex-sorts before its provider's. The retry loop closes the gap.

## Why this actually solves it

Verified end-to-end:

- **Boot path:** `Reg.LoadState` → `transitionState` → `runner.Transition` (`registry.go:557`). The fix is on the actual boot codepath.
- **Error chain:** every "X not found" rejection from the listener layer (queue, http, fs, template, aws.s3, aws.config) constructs an `apierror.Error` with `Kind = NotFound`. `isDeferrable` walks the unwrap chain via `errors.As` and matches them through `NewOperationRejectedError`'s wrap.
- **Retry safety:** the canonical failing listeners validate-before-mutate (queue consumer calls `GetQueue` / `GetDriver` before `m.consumers.Store` and `bus.Send(ServiceRegister)`), so re-running `Add` after a deferred failure produces no double-side-effects.

## Empirical proof

In-process MRE under `tests/proof/determinism/boot_loop_test.go` constructs two entries with no declared dependency between them — a provider whose listener `Add` registers a resource id, and a consumer whose `Add` returns `apierror.NotFound` if the resource isn't registered. Two name-order variants, N=2000 iterations each. Run via `bash tests/proof/run_pr270_proof.sh 2000`:

| | Variant A (provider lex first) | Variant B (consumer lex first) |
| --- | --- | --- |
| **pre-fix (`main`)** | 996/2000 pass | 1002/2000 pass |
| **post-fix (HEAD)** | **2000/2000 pass** | **2000/2000 pass** |

The Variant B post-fix cell going from `0/2000` (sort-only commit, the previous head of this branch) to `2000/2000` (this commit) is the decisive evidence that the retry loop is doing real work, not just freezing the flake into a deterministic constant.

## Tests

- `system/registry/runner/bus_runner_deferral_test.go` — 6 subtests for the loop: clean-pass, consumer-before-provider retry, non-NotFound immediate rollback, unresolvable returns `NewUnresolvedDependenciesError`, deeper chain of providers+consumers converges, `isDeferrable` walks the wrap chain.
- N=1000 shuffle-property tests for each sort site (in `*_determinism_test.go` files).
- `tests/proof/run_pr270_proof.sh` writes the verdict matrix to `tests/proof/pr270_proof_verdict.md` and exits non-zero if Variant B post-fix ever drops below `N/N`.

Local validation: `go test ./...` green, `golangci-lint run ./...` clean for changed files.

## Out of scope

- Per-kind audit of `Resolver.RegisterPattern` coverage. The self-heal loop makes the registry order-independent, so missing patterns now degrade boot performance (an extra pass or two) rather than breaking it.
- Supervisor parallel-`Start()` semantics (window W2). Not the production failure mode reported.

### DIFF
--- a/internal/interpolator/string_replacer.go
+++ b/internal/interpolator/string_replacer.go
@@ -36,7 +36,7 @@ func (r *replacer) replaceRecursive(val reflect.Value, ctx any) (any, error) {
 	}
 
 	switch val.Kind() {
-	case reflect.Ptr, reflect.Interface:
+	case reflect.Pointer, reflect.Interface:
 		if val.IsNil() {
 			return val.Interface(), nil
 		}
@@ -44,7 +44,7 @@ func (r *replacer) replaceRecursive(val reflect.Value, ctx any) (any, error) {
 		if err != nil {
 			return nil, err
 		}
-		if val.Kind() == reflect.Ptr {
+		if val.Kind() == reflect.Pointer {
 			ptr := reflect.New(val.Type().Elem())
 			ptr.Elem().Set(reflect.ValueOf(elem))
 			return ptr.Interface(), nil
--- a/internal/testutil/shuffle.go
+++ b/internal/testutil/shuffle.go
@@ -1,0 +1,19 @@
+// SPDX-License-Identifier: MPL-2.0
+
+// Package testutil provides shared test helpers used across the runtime
+// for determinism and shuffle-property style tests.
+package testutil
+
+import "math/rand/v2"
+
+// ShuffleSlice returns a copy of s shuffled with a deterministic PCG source
+// derived from seed. Calling ShuffleSlice with the same seed always yields the
+// same permutation, so it is safe to use in regression tests that need to
+// reproduce a specific input order.
+func ShuffleSlice[T any](s []T, seed uint64) []T {
+	out := make([]T, len(s))
+	copy(out, s)
+	r := rand.New(rand.NewPCG(seed, seed^0xA5A5A5A5A5A5A5A5)) //nolint:gosec // deterministic test shuffle, not security-sensitive
+	r.Shuffle(len(out), func(i, j int) { out[i], out[j] = out[j], out[i] })
+	return out
+}
--- a/internal/testutil/shuffle_test.go
+++ b/internal/testutil/shuffle_test.go
@@ -1,0 +1,46 @@
+// SPDX-License-Identifier: MPL-2.0
+
+package testutil
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestShuffleSlice_Reproducible(t *testing.T) {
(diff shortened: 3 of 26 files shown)


wolfy-j:
Wait, but arent it topo sorted anyway, it was no dependency chain?

skhaz:
You're right on both counts — topo sort is honored, and there's no chain in the graph for the failing pairs. The seam is the tie-break between unconstrained nodes: stableTopologicalOrder ties on input slice index, and that slice came from range map upstream, so the "tie" was effectively the process hash seed (measured 49/50 distinct orderings on the same 9 IDs).

That said, I'm second-guessing whether closing that seam actually removes the production flake or just makes it deterministic. A minimum reproducible example would settle it — I'll try to craft one (mock provider + consumer with a lazy fs.get/queue.get, no declared depends_on, boot loop pre/post fix) and report back before pushing this further.

PR #273, opened 2026-05-23
engine: fix runtime channel-anchored subscription leak + ephemeral router

wolfy-j:
## Summary

Root-causes and fixes a slow GB-scale memory leak in long-running Wippy
processes (observed in `kickside.core.jobs.job_worker` and
`kickside.core.scheduler.scheduler` after ~100k poll iterations) plus the
chain of secondary bugs it exposed once the channel layer was hardened.

The leak originated in `deliverMessage`: external delivery for a
subscribed topic called `Channel.Send(nil, value, nil)` which, on a full
buffer with no waiting receiver, pushed a `chanOp{task: nil}` onto the
channel's `sendq`. The fake blocked-sender entry was never woken because
there was no producer task to resume — under sustained producer pressure
`sendq` accumulated indefinitely.

Replacing that single call with a non-blocking variant required
straightening out the surrounding model: a per-process ephemeral channel
router for clock/timer/ticker (and future websocket / http stream / etc.)
fires that need epoch/generation-aware delivery, a clock dispatcher that
can stop firings by router (epoch, chID) coordinates, and a strict
end-to-end retry path in `Process.Step` so rendezvous (zero-buffer) topics
like `@pid/events` and `@pid/inbox` never silently drop a frame.

## What changed

**Channel layer**
- `Channel.TrySend` — non-blocking send that hands off to a waiting
  receiver, buffers if there is room, or reports `sent=false` without
  pushing a phantom blocked-sender. Used by all external (nil-sender)
  delivery paths.
- `deliverMessage` now routes external sends through `TrySend`. On miss
  the message stays in `messageQueue` and is re-delivered on the next
  flush — preserving mailbox order and rendezvous semantics. Closed-
  channel subs are removed so the queue can drain. Terminal payloads
  apply the close result so blocked receivers wake with `(nil, false)`.
- `Process.Step` final-flushes `messageQueue` before the outer loop
  breaks. `processChannelYields` can park new receivers inside the same
  iteration the earlier flush ran in; without the retry those messages
  would sit until the next external event resumed the process. This is
  what was breaking `app.test.process:events_cancel`.

**Ephemeral channel router (per-process)**
- New `TopicEphemeral = "@pid/route"` reserved topic. Frames carry an
  `EphemeralFrame{Source, Payloads, Epoch, ChID, Gen, HasValue, Close}`
  envelope. `routeEphemeralFrame` dispatches by `ChID` and drops stale
  frames (mismatched epoch or generation) without locking the recv path.
- `OverflowPolicy` (`Drop` / `Coalesce` / `Close`) declared per entry so
  clock-style signals can intentionally drop ticks while ordered streams
  close on overflow.
- `RegisterEphemeral`, `BumpEphemeralGen`, `SetEphemeralProducerStop`,
  `StopEphemeral`, `Abort`, `drainEphemeralChannels`, and `HasSubscriptions`
  integrate the router with `Step`, the scheduler, and process lifecycle.
- `epoch` is `atomic.Uint64`, `router` is `atomic.Pointer[ephemeralRouter]`,
  per-entry `producerStop` is `atomic.Pointer[stopFunc]`, all to keep
  `Abort` (scheduler goroutine) safe alongside `Step` (step goroutine).

**Clock dispatcher**
- Epoch / generation-aware: `TimerStartCmd` and `TickerStartCmd` carry
  `{GenRef, Build, Topic, PID, Duration, ChID, Epoch}` so fires can be
  routed to the originating router entry.
- New `TimerStopByChIDCmd` / `TickerStopByChIDCmd` for stop-by-router-ID.
- `reverseMap` + `pendingStops` with a 5-minute sweeper handle the
  start-then-stop race.

**Timer / ticker module migration**
- `time.after`, `time.timer`, `time.ticker` now produce
  `EphemeralFrame` payloads via the router. `SubscribeExisting` / topic-
  handler / `AddCleanup` paths removed.

**Workflow integration**
- `service/temporal/workflow/timer_manager.go` carries the new fields
  (`build`, `genRef`, `epoch`, `chID`) and implements
  `StopTimerByChID` / `StopTickerByChID`.

**Events module**
- `events.Subscription:close` now releases the frame-store cleanup slot,
  detaches the channel from the process subscription map, and applies the
  close result so blocked receivers wake. Long-running processes that
  open many subscriptions no longer accumulate dead cleanup closures.

## Coverage

- `runtime/lua/engine/subscription_cleanup_test.go` — 9 regression tests
  for the rendezvous handoff, queue-until-park retry, terminal close,
  closed-sub drain, unsubscribe wake, handler removal, and the strict
  single-Step batched delivery that catches the loop-early-exit bug.
- `runtime/lua/engine/channel_trysend_test.go` — `TrySend` semantics.
- `runtime/lua/engine/ephemeral_test.go`, `ephemeral_stress_test.go`,
  `ephemeral_bench_test.go` — router correctness, race coverage, hot-path
  benchmarks.
- `runtime/lua/modules/time/leak_test.go` — end-to-end leak proofs for
  `time.after`/`timer`/`ticker` under sustained polling.
- `system/clock/router_test.go` — dispatcher epoch/gen/tombstone behavior.

## Verification

- `go vet ./...` — clean.
- `golangci-lint --build-tags=race ./...` — **0 issues**.
- `go test -race -count=1 ./runtime/lua/engine/ ./runtime/lua/modules/process/ ./runtime/lua/modules/events/ ./runtime/lua/modules/time/ ./system/clock/ ./api/clock/ ./service/temporal/workflow/` — all green.
- `tests/app/test.sh` with `docker compose up temporal minio elasticmq socks5-proxy` — **381/381 passed** in 11.6s. Before the fix `app.test.process:events_cancel` hung and the runner SIGKILLed at 32s.
- Live deployment — RSS plateau at 524 MB over 80 minutes of 10ms scheduler polling, vs pre-fix 1.1 GB over 11 hours at 1s polling.

## Test plan

- [x] `go vet ./...` clean
- [x] `golangci-lint --build-tags=race ./...` 0 issues
- [x] `go test -race ./runtime/... ./system/clock/... ./api/clock/... ./service/temporal/workflow/` green
- [x] `tests/app/test.sh` 381/381 passed
- [x] Live RSS plateau on deployed wippy under sustained 10ms polling
- [x] Race detector clean across stress + bench suites


wolfy-j:
Redo, no extra "routers", wtf.

PR #274, opened 2026-05-24
http: make static mount updates idempotent

wolfy-j:
## Summary
- make root static mounts entry-owned upserts so registry updates for the same http.static entry do not collide with their existing path
- preserve conflicts when a different static entry claims an occupied mount path
- add route-manager, server-service, and manager-level regressions including same-path update, path move, conflict, and rollback

## Validation
- go test ./service/http
- go test ./api/service/http ./service/http
- go test ./...

### DIFF
--- a/service/http/manager_test.go
+++ b/service/http/manager_test.go
@@ -534,8 +534,48 @@ func TestManager_StaticOperations(t *testing.T) {
 	// Verify pending server rebuild
 	assert.True(t, manager.pending[serverID])
 
+	// Updating the same static entry must replace the entry-owned mount in place.
+	err = manager.Update(ctx, staticEntry)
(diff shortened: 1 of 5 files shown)


wolfy-j:
Local validation from clean worktree /tmp/wippy-http-static-upsert:\n\n- go test ./service/http ✅\n- go test ./api/service/http ./service/http ✅\n- go test ./... ✅\n- make lint ✅\n- SKIP_TEMPORAL_TESTS=1 SKIP_CLOUDSTORAGE_TESTS=1 SKIP_DOCKER_TESTS=1 make test ✅\n\nGitHub Actions currently reports immediate Lint/Test failures without useful failed-step logs; local validation is green.

PR #305, opened 2026-06-03
fix(logs): restore console output dropped by per-emission metric hook

wolfy-j:
## Problem

Running `wippy run` (or `-c`) prints the banner + `initializing runtime` and then **goes silent** — no `infrastructure initialized`, `components loaded`, `runtime ready`, nothing. It looks like the runtime hangs/deadlocks at boot.

It does not. A goroutine dump shows the runtime fully booted (main parked in `waitForShutdownSignal`, ~1k healthy goroutines, gateway bound, `/health` → 200). The boot logs are simply being **dropped**.

## Root cause

`cmd/internal/logger` builds the CLI logger with `zapCfg.Build(zap.Hooks(syslogs.EmissionHook))` (added in #241 for `runtime_log_emissions_total`). `zap.Hooks` wraps the core in a zapcore `*hooked` core.

`boot.NewBootstrapContext` then wraps that logger with the event-streaming `system/logs.Core`, taking the CLI logger's core as its **downstream** and propagating via `downstream.Write(ent, fields)` directly.

But zapcore's `hooked.Write` only runs the hook funcs — it does **not** write to its underlying console core:

```go
func (h *hooked) Write(ent Entry, _ []Field) error {
    // Since our downstream had a chance to register itself directly with the
    // CheckedMessage, we don't need to call it here.
    var err error
    for i := range h.funcs { err = multierr.Append(err, h.funcs[i](ent)) }
    return err
}
```

It relies on the `Check` path (where the real core registers itself with the `CheckedEntry`). `logs.Core` bypasses that and calls `Write` directly, so once the runtime switches to the wrapped logger (right after `initializing runtime`) **all console output is silently dropped**.

## Fix

Remove the per-emission log metric hook so the CLI logger's core is a plain console core again (which `downstream.Write()` emits correctly):

- drop the `zap.Hooks` wiring in `cmd/internal/logger/logger.go`
- delete `system/logs/emissionhook.go` (+ test)
- drop `SetEmissionCollector` calls in `boot/components/metrics/metrics.go`

(Alternative considered: make `logs.Core.Write` delegate through `downstream.Check(...).Write(...)` to preserve the metric. Removed instead since the per-emission counter wasn't worth a wrapper-core contract this fragile.)

## Verification

`wippy run` now prints the full boot sequence through `runtime ready`; `/health` → 200. `go build`, `go vet`, and `go test ./system/logs/...` pass.


skhaz (approved):
Looks good to me.

PR #306, opened 2026-06-03
fix(eventual): gossip-replicate registry deltas to all members

skhaz:
Registry EVENTUAL-scope deltas were not reaching client (non-member) nodes: the origin emitted each delta once to a few peers and receivers did not re-broadcast, so convergence on client nodes took ~38s.

## Fix
- Epidemic re-broadcast of incoming deltas.
- Peer-authoritative `MergeRemoteState`.
- `PushPullInterval` 15s -> 5s.

Convergence drops to sub-second and `node_client` receives names directly.

## Proof
Full `go test ./... -race` green (286 packages, 0 failures/races). Covered by `cluster/membership` and `system/topology/namereg/eventual` suites (incl. `epidemic_test`).

### DIFF
--- a/cluster/membership/eventual_integration_test.go
+++ b/cluster/membership/eventual_integration_test.go
@@ -1,0 +1,211 @@
+// SPDX-License-Identifier: MPL-2.0
+
+package membership
+
+import (
+	"context"
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/wippyai/runtime/api/pid"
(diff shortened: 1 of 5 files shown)


skhaz:
Superseded by #307, which already contains this exact fix (commit 6268eea4 — gossip-replicate eventual registry deltas) plus an added client-publish / bidirectional test proving eventual name reg works across non-raft (role=client) nodes. Closing to avoid a duplicate change landing on main. #307 CI is green.

PR #345, opened 2026-06-15
fix(cmd): keep app context alive for graceful pack shutdown

skhaz:
## Why
`wippy run <hub-module>` (and `.wapp` pack files) logged
`failed to stop controllers during shutdown: supervisor is stopped: context canceled`
on exit and abandoned services instead of stopping them gracefully.

## What
The pack runner (`runPackEntries`) passed the `appCtx` cancel func as the
first-signal callback, so the shutdown signal cancelled the context the
supervisor and all its controllers derive from **before** `shutdown.Perform`
ran the graceful stop — making every `controller.Stop()` short-circuit. Pass
`nil` instead, matching the server path in `run.go`. The existing
`defer cancel()` still tears down `appCtx` after shutdown completes.

## Testing
- `make test` — 292 packages, 0 failures; `golangci-lint` v2.8.0 — 0 issues.
- New regression test `TestRunPackEntries_GracefulShutdownStopsRunningService`
  (`-race`, `-count=5`, no flakes): fails when reverted to `cancel`, passes with `nil`.
- Gremlins mutation on `run_pack.go`: 36 killed, 0 lived, 100% efficacy.
- Manual repro on three hub modules (`kickside/kickside`, `keeper/keeper`,
  `userspace/users`): shutdown "supervisor is stopped" errors 3 → 0; services
  now log clean graceful stop.

### DIFF
--- a/cmd/wippy/cmd/run_pack.go
+++ b/cmd/wippy/cmd/run_pack.go
@@ -494,7 +494,7 @@ func runPackEntries(
 		}
 	}
 
-	waitForShutdownSignal(sigChan, logger, cancel)
+	waitForShutdownSignal(sigChan, logger, nil)
 
 	exitCode := shutdown.Perform(ctx, loader, logger, silentLogs)
 	if exitCode != 0 {
--- a/api/supervisor/shutdown.go
+++ b/api/supervisor/shutdown.go
@@ -35,6 +35,8 @@ func GetExitCode() int {
 func SetSignalChannel(ctx context.Context, ch chan<- os.Signal) {
 	ac := ctxapi.AppFromContext(ctx)
 	if ac != nil {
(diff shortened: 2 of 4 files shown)


cmd/wippy/cmd/run_pack.go
```diff
@@ -431,7 +431,7 @@ func runPackEntries(
 		}
 	}
 
-	waitForShutdownSignal(sigChan, logger, cancel)
+	waitForShutdownSignal(sigChan, logger, nil)
```
skhaz, line 497: Don't cancel appCtx on the signal — it pre-empts the graceful stop. Let shutdown.Perform run first; defer cancel() cleans up after.

wolfy-j (approved):

PR #352, opened 2026-06-16
fix(temporal): resolve boundary type names as strings (no leading colon for namespace-less)

skhaz:
## Why

Spawning a workflow whose registry ID has no namespace (e.g. an external, PHP-hosted workflow) recorded the Temporal `WorkflowType` as `":PopulateItemRelation"` instead of `"PopulateItemRelation"`. Temporal type names are plain strings that only *sometimes* coincide with a registry ID; the runtime derived them by stringifying `registry.ID` ad-hoc across the dispatch paths, so a namespace-less ID rendered in its canonical `:name` form leaked into the type name. There was also no first-class way to name an external workflow.

## What

- **Boundary names are plain strings.** A single resolver (`service/temporal/options/name.go`) owns the rule: prefer an explicit name, else the qualified id (`name` when the namespace is empty, `ns:name` otherwise — never a leading colon). The qualify logic lives entirely in the temporal options package; `registry.ID` is untouched.
- **Explicit name controls.** New `workflow.name` / `activity.name` spawn options (with `temporal.*` legacy aliases) and a `MetaActivityName` registration override mirroring the existing `MetaWorkflowName`.
- **All dispatch + registration sites** (host spawn, child spawn/exec, workflow exec, activity dispatch, continue-as-new, listeners) resolve through the resolver. `registry.ID` stays the carrier on `Start.Source` (the local process path needs it as a key) but is demoted to a naming fallback.
- **Application-level tests** under `tests/app` exercise the real Lua surface: a workflow registered under an explicit string name, spawned by a bare custom name and via the `workflow.name` option override.
- **Incidental fix:** narrowed `Message?` -> `Message` at 13 call sites in 5 pre-existing temporal app fixtures (latent Luau type error that blocked the temporal app suite from loading; unrelated to naming).

## Testing

- `make build-wippy` + `go build ./...` green; `golangci-lint` 0 issues; `go test -race ./service/temporal/... ./api/registry/` green.
- Go unit/integration: resolver precedence/fallback (canonical vs legacy alias, empty/non-string/nil -> fallback, namespace-less -> no colon); host dispatch wire-capture (`workflow.name` override and namespace-less default -> `PopulateItemRelation`); workflow + activity listener register/unregister symmetry with a custom name. gremlins on the resolver: 100% killed.
- **Application surface against a real Temporal dev server: app suite 423/423, temporal suite 41/41** including the new `workflow_custom_name` test (bare-name dispatch + option override end-to-end).

### DIFF
--- a/api/registry/id.go
+++ b/api/registry/id.go
@@ -37,17 +37,6 @@ func (t *ID) String() string {
 	return t.NS + ":" + t.Name
 }
 
-// Qualified returns the namespace-qualified identifier as "ns:name", or just
-// "name" when the namespace is empty. Unlike String, it never emits a leading
-// colon, making it suitable as an external type identifier such as a Temporal
-// workflow or activity type name.
-func (t *ID) Qualified() string {
-	if t.NS == "" {
-		return t.Name
-	}
-	return t.String()
-}
-
 // Equal returns true if both IDs have the same namespace and name.
 func (t *ID) Equal(other ID) bool {
 	return t.NS == other.NS && t.Name == other.Name
--- a/api/registry/id.go
+++ b/api/registry/id.go
@@ -37,6 +37,17 @@ func (t *ID) String() string {
 	return t.NS + ":" + t.Name
 }
 
+// Qualified returns the namespace-qualified identifier as "ns:name", or just
+// "name" when the namespace is empty. Unlike String, it never emits a leading
+// colon, making it suitable as an external type identifier such as a Temporal
+// workflow or activity type name.
+func (t *ID) Qualified() string {
+	if t.NS == "" {
+		return t.Name
+	}
+	return t.String()
+}
+
 // Equal returns true if both IDs have the same namespace and name.
 func (t *ID) Equal(other ID) bool {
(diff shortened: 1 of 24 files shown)


api/registry/id.go
```diff
@@ -37,6 +37,17 @@ func (t *ID) String() string {
 	return t.NS + ":" + t.Name
 }
 
+// Qualified returns the namespace-qualified identifier as "ns:name", or just
+// "name" when the namespace is empty. Unlike String, it never emits a leading
+// colon, making it suitable as an external type identifier such as a Temporal
+// workflow or activity type name.
+func (t *ID) Qualified() string {
```
wolfy-j: can you drop it?

wolfy-j:
looks good

wolfy-j (approved):

api/registry/id.go
skhaz: Done — dropped in b7ac22bc. `api/registry/id.go` is back to its original form; the qualify logic now lives as an unexported helper in `service/temporal/options/name.go`, so `registry.ID` stays untouched.

PR #369, opened 2026-06-19
fix(wasm): bound warm-instance linear memory by recycling

skhaz:
## Why
A reused warm WASM component instance (per inline worker) keeps its linear memory alive across calls; WASM linear memory only ever grows, so each conversion's guest allocations accumulate **without bound**. Under sustained document conversion, RSS climbed past **6 GB** and the Go heap past **2 GB** (~62 KB retained per call for an image-heavy PDF).

## What
Recycle the warm instance after `recycleWarmInstanceAfter` (512) synchronous calls: close it so the next call instantiates a fresh one and reclaims the linear memory, while keeping the fast warm path for the common case. Re-instantiation is now safe because `wasm-runtime` is bumped to a version that **rebuilds the per-instance import bridges**, so a fresh instance no longer reuses a bridge bound to a closed instance's core.

- `runtime/wasm/engine/process.go`: `warmCalls` counter + recycle in `stepSync`.
- `go.mod`: bump `wasm-runtime` → `cc678f6` (per-instance bridge rebuild).
- regression test `TestProcessRecyclesWarmInstanceAfterThreshold`.

## Validation (10,000× sustained convert loop)
| | before | after |
|---|---|---|
| RSS | → ~6 GB | **stable ~280 MB** |
| Go heap (`alloc`) | → ~2.2 GB | **38–57 MB band** |
| `heap_objects` | — | **~flat** |
| conversions | trap on 2nd+ | **10000/10000 ok** |

`golangci-lint` clean; full `./runtime/wasm/...` suite green under `-race`; full `make test-runtime` green; independent code review verdict: ship.

## Follow-up (not in this PR)
Re-instantiation recompiles the synthetic bridge and its wazero `CompiledModule` is not freed on `Close` (~42 objects per re-instantiation). Recycling amortizes this to ~0.07 obj/call (negligible), but a complete fix would free the bridge `CompiledModule` on `Close` in `wasm-runtime`. `recycleWarmInstanceAfter` (512) could be made env-tunable.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -66,7 +66,7 @@ require (
 	github.com/wippyai/tree-sitter-markdown v0.0.3
 	github.com/wippyai/tree-sitter-sql v0.0.4
 	github.com/wippyai/wapp v0.1.2
-	github.com/wippyai/wasm-runtime v0.0.0-20260209224309-586ea6933075
+	github.com/wippyai/wasm-runtime v0.0.0-20260619200926-cc678f6c6549
 	github.com/xuri/excelize/v2 v2.10.1
 	go.opentelemetry.io/otel v1.44.0
 	go.opentelemetry.io/otel/exporters/otlp/otlpmetric/otlpmetricgrpc v1.40.0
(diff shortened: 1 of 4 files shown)


wolfy-j (changes requested):
need better api

wolfy-j:
handled with proper post-call sentiel error

PR #391, opened 2026-06-24
fix(lua): guard cyclic table conversion

wolfy-j:
## Summary
- Prevent Lua table conversion from recursing forever on cyclic tables by tracking the active conversion stack.
- Preserve existing acyclic table conversion behavior; cyclic references become a bounded sentinel string.
- Add unit coverage for cyclic map and array tables.
- Add an app-level process context regression so tests/app/test.sh covers the runtime boundary that previously hit TableToMap -> ToGoAny recursion.

## Validation
- go test ./runtime/lua/engine/value -count=1
- tests/app runner passed in the existing runtime checkout with optional external suites skipped: 385 tests

Note: running the app test runner from a clean origin/main worktree currently stops before tests on an unrelated eval config decode issue: app.test.eval:runner_allow_classes has security as an array while main expects security.Config.

### DIFF
--- a/runtime/lua/engine/value/util.go
+++ b/runtime/lua/engine/value/util.go
@@ -252,8 +252,14 @@ func GetFunc(l *lua.LState, value lua.LValue, field string) (*lua.LFunction, boo
 	return nil, false
 }
 
+const cyclicTableSentinel = "<cyclic table>"
+
 // ToGoAny converts a lua.LValue to its Go equivalent.
 func ToGoAny(v lua.LValue) any {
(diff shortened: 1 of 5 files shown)


wolfy-j:
Superseded by #390, which fixes the same cyclic Lua table conversion bug. Both guard `ToGoAny`/`TableToMap`/`TableToSlice` against infinite recursion with equivalent ancestor-based cycle detection; #390 was chosen because its path-slice uses a stack-allocated array (zero heap allocation for the common shallow/acyclic case) on this hot conversion path, whereas this PR allocates a map per call. Thanks for the fix here, including the process-context integration fixture.

PR #414, opened 2026-07-02
feat: support for environment-based configuration via `OptionsEnv`.

VladKhomich:
# Reason for This PR

A `db.sql.postgres` resource can already read its host, port, database, username, and password from environment variables (host_env, port_env, and so on), but the connection options map can only be hardcoded. `sslmode` lives in that map, and that's a problem when the same config has to run against databases with different TLS setups: production RDS enforces TLS, while local and CI Postgres have none.

Until now the only ways around it were editing the committed config by hand for local runs or passing an `-o` `app:db:options.sslmode=...` override on every command. Both are easy to forget and easy to get wrong. This change lets connection options come from the environment the same way the other fields do.

## Description of Changes

- Added an `options_env` field to `DBConfig`. It maps an option name to an environment variable, e.g. `sslmode: DB_SSLMODE`, so `sslmode` (default require) can be flipped to disable per environment without touching the config.
- `resolveDBConfigEnv` now resolves those bindings into the options map. An env-provided value wins over a static one for the same key. Values are resolved into a staging map and only merged once all lookups succeed, so a single failed lookup can't leave options half-populated; the error identifies which option and variable failed.
- Validate rejects entries with an empty option name or an empty variable reference, and `InitDefaults` initializes the map next to the existing options map.
- Tests cover the override, non-overlapping keys, a nil options map, multiple bindings applied together, empty/invalid entries, and the all-or-nothing failure behavior.


wolfy-j (approved):

wolfy-j:
Closing in favor of #418, which generalizes exactly this. That PR centralizes all `_env` resolution into the config decoder, so connection options resolve from the environment for every service — not just SQL — via the suffix convention inside the options map: `options: { sslmode_env: DB_SSLMODE }` (an env value wins over a static one for the same key, and a single failed lookup fails the decode rather than half-populating). 

The dedicated `options_env` field and the extended `resolveDBConfigEnv` here are **superseded** — #418 removes the per-service `_env` fields and resolvers entirely while keeping old `*_env:` configs working. Your `sslmode` per-environment use case is covered end to end there. Thanks for the work and for surfacing the gap — it's what drove the broader cleanup.

PR #415, opened 2026-07-02
fix(wasm): deliver preview1 WASI env/mounts to core modules

skhaz:
## Why
A core `wasi_snapshot_preview1` `function.wasm` (e.g. an R interpreter driven from Lua) booted with **no env vars and no preopened dirs**, aborting at startup (`R home directory is not defined`, then `cannot create R_TempDir`). Env vars declared only via `default:` never resolved either.

## What
- Thread the resolved WASI args/env/mounts into the engine `InstanceConfig` (wazero's preview1 host reads env/preopens from the module config).
- Resolve env via `Get` (default-aware) instead of `Lookup` (storage-only).
- Mount host dirs via `RootPath` so writable mounts (a scratch `/tmp`) actually allow writes.
- Bump `wasm-runtime` to the branch adding `list<u8>` WIT parsing + reactor `_initialize`.

## Testing
- `golangci-lint` clean; build green.
- `runtime/wasm/...` 13 pkgs ok; full WASM app suite **388 tests pass**, incl. wasm group **9/9** (js/fs/http/process/mapper/complex/sleep/wat) — nothing regressed.
- A real **R 4.6.1** reactor now runs end-to-end via `funcs.call`, including `r.eval(code, luaArgs...)`.

**Depends on** wippyai/wasm-runtime#10 (merge first; the go.mod bump points at that branch commit and should be re-pinned to the merged version).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


runtime/wasm/engine/process.go
```diff
@@ -381,7 +399,9 @@ func (p *Process) resolveWASICallConfig(ctx context.Context) (*wippyhost.WASICal
 		callCfg.Env = make(map[string]string, len(p.wasi.Env))
 		for _, item := range p.wasi.Env {
 			id := item.ID.String()
-			value, found, err := envReg.Lookup(ctx, id)
+			// Get (not Lookup) so a variable with only a default value still
+			// resolves — Lookup reads storage only and misses defaults.
+			value, err := envReg.Get(ctx, id)
```
wolfy-j, line 408: must respect security boundaries

runtime/wasm/engine/process.go
```diff
@@ -229,10 +229,28 @@ func (p *Process) startExecution() error {
...
+		// Core wasi_snapshot_preview1 modules read env/args/preopens from the wazero
+		// module config, so thread the resolved WASI mapping through.
+		if wc := wippyhost.GetWASICallConfig(execCtx); wc != nil {
+			instCfg.Args = wc.Args
+			instCfg.Env = wc.Env
+			for _, mnt := range wc.Mounts {
+				m := wasmengine.Mount{Guest: mnt.Guest, ReadOnly: mnt.ReadOnly}
+				// Prefer a host directory path (supports read-write, needed for e.g.
+				// a scratch /tmp); fall back to a read-only fs.FS mount otherwise.
+				if rp, ok := mnt.Filesystem.(interface{ RootPath() string }); ok {
```
wolfy-j: not sure i like this interface here

runtime/wasm/engine/process.go
```diff
@@ -229,10 +229,28 @@ func (p *Process) startExecution() error {
...
+					m.Host = rp.RootPath()
+				} else {
+					m.FS = mnt.Filesystem
+				}
+				instCfg.Mounts = append(instCfg.Mounts, m)
```
wolfy-j, line 251: does it mean all wasm always have access to root fs?

runtime/wasm/engine/process.go
skhaz: Done in b37c546e — WASI env resolution now gates each read with `security.IsAllowed(ctx, "env.get", id, nil)` (same action the Lua env module uses) and denies with a dedicated `NewWASIEnvAccessDeniedError` otherwise.

runtime/wasm/engine/process.go
skhaz: Fixed in b37c546e — replaced the inline anonymous interface with a named `fsapi.HostPathFS` (in `api/fs`), and moved the resolution into `resolveWASICallConfig` (right after the `fs.get` access check) so the engine no longer duck-types it — it just consumes the resolved `Host`/`Filesystem` on the binding.

runtime/wasm/engine/process.go
skhaz: No — a guest never gets the host root. Each mount is served from its own filesystem's sandboxed root path (`fsapi.HostPathFS.RootPath()`) and wazero re-roots the guest at exactly that directory. A guest only ever sees the specific directories declared in its `wasi.mounts`, and every one is already gated by `security.IsAllowed(ctx, "fs.get", fsID)`. Clarified in b37c546e. (Follow-up idea per your other note: a per-call dir-prefix via `funcs.with_options()` would slot in the same way `network.select` does, since WASICallConfig is already per-invocation.)

PR #416, opened 2026-07-03
feature: migrate command

VladKhomich:
## Reason for This PR

There is no way to run database migrations on their own. Today migrations only run as a side effect of booting the whole application: the bootloader service discovers the migration bootloader during startup and applies pending migrations before the app comes up.

That is fine for a normal boot, but it leaves a gap for everything else. CI pipelines, one-off ops tasks, and local resets all want the same thing that Entity Framework (dotnet ef database update), Prisma (prisma migrate deploy), and Alembic already give you: apply the migrations, print what happened, exit with a status code.

## Description of Changes

`wippy migrate up` and `wippy migrate down --count N`, both applying to every target database found in the registry. Verified end to end against a fresh Postgres: up applied 47 migrations, down --count 3 rolled back the three newest, and a second up re-applied them, all with a clean exit code.

`[Author TODO: add description of changes.]`

## Possible follow-ups
`wippy migrate status` to list applied and pending migrations, backed by the runner's existing status function (another small module entry).
A `--database` flag to target a single database instead of all of them.
`--dry-run` and `--json` for CI use.

### DIFF
--- a/cmd/wippy/cmd/errors.go
+++ b/cmd/wippy/cmd/errors.go
@@ -197,3 +197,19 @@ func NewInstallFailedError(cause error) apierror.Error {
 func NewLintFailedError(errors, warnings int) apierror.Error {
 	return apierror.New(apierror.Invalid, fmt.Sprintf("lint failed: %d errors, %d warnings", errors, warnings)).WithRetryable(apierror.False)
 }
+
+func NewFunctionRegistryNotAvailableError() apierror.Error {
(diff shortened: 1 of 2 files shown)


skhaz (approved):

wolfy-j:
Closing: migrations are an application-layer concern, not part of the runtime. The migrate command belongs in the app rather than in the runtime repo. We can model it as `wippy run NAME` where name is command like `test`.

PR #422, opened 2026-07-04
refactor(runtime): generic frame-context resolver registry

wolfy-j:
## Why

Options that decorate a newly-spawned frame's context (today: the network overlay; soon: a filesystem `fs_root`) were hand-wired into **every** dispatcher — `system/process/manager.go` `Start()` and both `runtime/{wasm,lua}/component/function/lifecycle.go` `Execute()` each called `netapi.ApplyOverlayPair`. That leaks `api/net` (and would leak `api/fs`, `api/service/otel`, ...) into generic machinery, and every new frame option must be added to every dispatcher.

## What

`ctxapi.FrameResolvers` — an ordered registry of resolvers:
```go
type FrameResolver func(ctx, options attrs.Attributes) ([]Pair, error)
```
Carried on the AppContext (like the interceptor registry), populated once at boot. The two real choke points apply the whole set generically:
- `system/function/functions.go` `executor()` — shared by lua **and** wasm functions.
- `system/process/manager.go` `Start()`.

Adding a frame-context option is now **one resolver + one `Register` line**, zero dispatcher edits. The dispatchers no longer import `api/net`/`api/fs`.

### Network migrated
`api/net.OverlayResolver()` replaces `ApplyOverlayPair` (same logic: per-call → app-default precedence, `ErrNetworkNotFound` for unknown ids) and is registered by the network boot component (`order 200`). The `netapi` calls + imports are removed from all three dispatchers.

## Performance
Reads are **lock-free** — an atomic snapshot pointer (copy-on-write under a mutex only in `Register`), mirroring the interceptor registry's `chain.Load()`. Benchmarks:
- no resolvers (the common no-overlay call): **1.4 ns/op, 0 allocs** — just the atomic load.
- 2 resolvers: 145 ns/op — the same per-resolver work as the old direct call.

Race-tested (`go test -race`).

## Not in this PR
- `fs_root` registers its resolver on this mechanism (that PR builds on this one).
- otel span propagation lives in the otel interceptor (not a dispatcher leak); it can migrate to a resolver as a no-risk follow-up.

## Tests
- `api/context`: registry order, append-to-input, nil-receiver no-op, first-error-stops-and-wraps (`errors.Is` preserved), duplicate/nil rejection, ctx round-trip, benchmarks.
- `api/net`: `ApplyOverlayPair` tests retargeted to `OverlayResolver`.
- `system/process`: manager network tests now wire the resolver registry, proving the manager applies it generically.


wolfy-j:
Superseded by #420. The frame-context resolver commit stack has been moved onto feat/wasm-fs-prefix-option so #420 is now the resolver PR.

PR #423, opened 2026-07-05
deps: install pipeline hardening — link-stage fan-out, diamond fixpoint, upgrade signaling, env admission

wolfy-j:
# deps: install pipeline hardening — link-stage fan-out, diamond fixpoint, upgrade signaling, env admission

## Summary

Hardens the dependency install/uninstall pipeline around PR #423. The link stage, changeset expansion, and admission policies now handle transitive graphs, parameterized roots, and repeated re-resolution reliably; the full install → update → uninstall cycle is proven end to end against a real application (kickside) and the public hub.

### Link stage (`boot/build/stages/linking.go`)
- **Bare-name fan-out restored as the linking convention**: a bare dependency parameter fans its value out to every owned requirement sharing that name across sub-namespaces; a fully-qualified id addresses exactly one. The only error is a genuine value conflict on the same concrete requirement (root-over-transitive precedence preserved). The interim eager-ambiguity rejection is removed — it broke applications that rely on the fan-out convention.
- **Diamond re-resolution fixpoint**: worklist-based normalization so diamond-shaped graphs converge instead of failing on order-dependent conflicts.

### Changeset expansion (`boot/deps/hub`)
- Deleting a dependency root removes the module's `ns.definition` and all module-owned edges in the same atomic DirectiveResult; re-expansion never re-admits a deleted module's entries with default parameters. Locked in by faithful repro tests (`dependency_handler_uninstall_repro_test.go`) modeling a parameterized module with conflicting default routes — self-delete and unrelated-delete both emit exact deletions and preserve sibling parameters.

### Process upgrades
- `upgradable` option + `OUTDATED` notification on code invalidation, so supervisors can `process.upgrade` instead of being torn down.

### Admission
- Legacy `*_env` directives keep their inline value when the name is not a registered variable instead of failing admission.

### Tooling
- `test.sh` now covers `./boot/...` and `./system/registry/...` — the packages this PR touches — alongside the existing suites.

## Testing
- `go test ./boot/... ./system/registry/...` green; new coverage for module-qualified fan-out, three-way and mixed addressing, diamond convergence, and the parameterized-uninstall delete path.
- Live E2E against a real app + public hub: 10 installs (incl. depth-4 shared chains), pinned-version install → update-in-place, uninstalls with shared-transitive retention, parameterized root install → uninstall cycle clean.

## Compatibility
- Restores (not changes) the established fan-out linking convention; applications written against it boot unchanged. Value conflicts still fail loudly.


wolfy-j:
wip

skhaz (approved):

wolfy-j:
Hardening series landed on this branch (4c83856f4..52c8ddf29): link-stage module-qualified fan-out, diamond constraint-intersection fixpoint, OUTDATED upgrade signaling, legacy *_env admission, lock fallback scoped to installed roots, plus test.sh coverage for boot and system/registry. go test ./boot/... ./system/registry/... green.

PR #438, opened 2026-07-06
fix(linker): keep requirement parameters namespace-owned

wolfy-j:
## Summary
- scope dependency ownership to the component namespace and its child namespaces
- keep legacy one-segment meta.module aliases working
- prevent unrelated dotted namespaces from accepting foreign bare parameters via meta.module

## Why
The app hit api_router conflicts because a dotted requirement namespace like wippy.views could still be treated as owned by another module through meta.module. That let keeper/keeper and kickside agents parameters collide with the real wippy/views parameter.

## Tests
- go test ./boot/build/stages
- go test ./boot/deps/hub

### DIFF
--- a/boot/build/stages/linking.go
+++ b/boot/build/stages/linking.go
@@ -240,15 +240,23 @@ type decodedDependency struct {
 	transitive bool
 }
 
-// owns reports whether a dependency addresses a requirement: the requirement
-// lives in the dependency component's module namespace, or its meta.module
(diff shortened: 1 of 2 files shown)


skhaz (approved):

wolfy-j:
Superseded by #437, which carries the linker requirement-scope fix plus the loader/package-manifest hardening and updated tests.

PR #448, opened 2026-07-07
feat(lua): Python-style traceback + locals capture

skhaz:
Python-style stack-trace + locals capture for Wippy Lua scripts, on demand.

## Why
Lua scripts had no way to capture a stack trace with local variables — only a bare `debug.traceback` string with no locals, and even that was unavailable because `SkipOpenLibs` excluded the `debug` library entirely.

## What
- **Read-only `debug` subset** as a cached global (`cached_libs.go`): `traceback`, `getinfo`, `getlocal`, `getupvalue`. Stripped: `setlocal`, `setmetatable`, `setupvalue`, and `getmetatable` (the last bypasses `__metatable` — a sandbox hole).
- **`traceback` module** (`runtime/lua/modules/traceback`):
  - `traceback.format(level, opts)` — formatted multi-line string (Python `traceback.format_exc` style).
  - `traceback.frames(level, opts)` — structured frames with `locals` and `upvalues`.
  - Options: `depth`, `locals`, `upvalues`, `max_value_len`, `max_table_depth`.
  - Cycle detection for self-referential tables; frames that panic under post-recover conditions are skipped instead of aborting the trace.
- Registered as a core ambient module (`core_modules.go`) — `require("traceback")` and global.

## Usage
```lua
-- On-demand snapshot
print(traceback.format(1))

-- At the throw site (xpcall handler runs with stack intact)
xpcall(risky, function(e) return traceback.format(1) end)
```

## Dependencies
Bumps `github.com/wippyai/go-lua` to a pre-release commit (wippyai/go-lua#32) for two prerequisites: `xpcall` catching errors under direct `PCall`/`DoString`, and `inspect.GetStackFrame` capturing upvalues. Once wippyai/go-lua#32 merges and tags a release, the pseudo-version can be replaced with the tag.

## Testing
- TDD: RED → GREEN for the module, the cached-lib subset, and both fork fixes.
- `make test-runtime`, `make lint`, gofmt, vet — green.
- Mutation: 65% efficacy, 100% coverage; correctness-critical mutants killed.

## Sample output (xpcall throw-site capture)
```
[3] <string>:6 (risky)
    Locals:
      depth = 0
      context = processing layer 0
      payload = {secret = shh, id = 0}
    Upvalues:
      risky = <function>
```

No visual/UI changes (runtime-only feature).

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -63,7 +63,7 @@ require (
 	github.com/tree-sitter/tree-sitter-php v0.24.2
 	github.com/tree-sitter/tree-sitter-python v0.25.0
 	github.com/tree-sitter/tree-sitter-typescript v0.23.2
-	github.com/wippyai/go-lua v1.5.16
+	github.com/wippyai/go-lua v1.5.17-0.20260707180131-68784b1e34c8
 	github.com/wippyai/module-registry-proto-go v0.0.1
 	github.com/wippyai/tree-sitter-markdown v0.0.3
 	github.com/wippyai/tree-sitter-sql v0.0.4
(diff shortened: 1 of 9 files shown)


wolfy-j:
will be kept for 2.0.0 update of go lua since it already has some pcall work

wolfy-j (changes requested):
Blocking release review:\n\n1. This depends on wippyai/go-lua#32, which is still open, via a pseudo-version. Merge and tag that prerequisite before Runtime consumes it.\n2. The claimed read-only surface is not actually read-only: `debug.getlocal`, `debug.getupvalue`, and `traceback.frames` return live table/userdata references from caller frames. A callee can mutate a caller/upvalue table even though the setter functions were removed. Keep introspection internal and return bounded snapshots instead.\n3. Formatting is not memory-bounded. `max_value_len` caps individual strings and `max_table_depth` caps depth, but table width and total output are unlimited; one wide local table can exhaust the process. Add entry and total-byte caps and tests.\n\nThe feature direction is useful, but it should not ship as an ambient debug capability with mutable references and unbounded output.

PR #461, opened 2026-07-09
feat(config): compose runtime config files

skhaz:
## Why

Runtime environments need to compose settings without mutating the shared project baseline or inventing reserved local filenames. Workspace checkout paths are runtime/environment state, while `wippy.lock` remains the portable exact dependency snapshot.

## What

`--config` is now repeatable on runtime commands:

```sh
wippy run \
  --config .wippy.yaml \
  --config .wippy.dev.yaml \
  --config config/postgres-history.yaml \
  --config .wippy.workspace.yaml \
  --profile workspace
```

- Files merge left to right; later files override matching leaves and preserve unrelated settings.
- Every explicitly selected file must exist. With no `--config`, `.wippy.yaml` remains the optional default.
- The first file remains the project/config-directory anchor for relative runtime paths. The complete absolute list is exposed as `boot.config_paths`.
- Profiles apply after file composition, in requested order. CLI overrides such as `--set` apply last.
- Filenames have no reserved meaning, no overlay is auto-discovered, and `wippy init` does not modify ignore rules.
- Logger preloading uses the same ordered stack, so bootstrap logging and runtime boot cannot disagree.

Any selected config can declare environment state, including multiple workspace replacements:

```yaml
version: "1.0"

profiles:
  workspace:
    workspace:
      replacements:
        wippy/runtime: ../runtime
        acme/ui: ../ui
```

`wippy run`, `install`, `update`, `lint`, registry commands, and packing all consume the same composition path.

## Dependency and history behavior

- `wippy.lock` remains the portable exact module/version snapshot. Effective workspace replacements alter the load graph but are never serialized into the lock.
- Existing tracked lock replacements remain readable for compatibility and emit a deprecation warning directing users to runtime configuration.
- Replacement paths resolve relative to the first runtime config and must identify directories.
- The dependency handler snapshots effective lock/replacement state at boot. Replacement trees are rehashed before history reuse, preventing changed or missing local sources from being mistaken for a stored baseline.

## Publishing safeguards

- The runtime `--config` stack is not a publishing input; `publish --config` remains the module-manifest directory option.
- `workspace` configuration is unconditionally removed from published runtime metadata.
- `publish.profiles.include` explicitly selects publishable non-workspace profiles. `include: []` publishes none; an unknown profile fails.

```yaml
publish:
  profiles:
    include: [production]
```

For compatibility, omitting `include` keeps the previous behavior for non-workspace profiles only.

## Validation

- `go test ./...`: pass.
- Focused race tests for config loading and CLI composition: pass.
- `make lint`: clean.
- Production binary build: pass.
- Binary smoke: three arbitrary config files + profile + final `--set` boot successfully; a missing explicit file fails immediately.
- Publishing isolation and workspace-stripping regression tests: pass.

### DIFF
--- a/README.md
+++ b/README.md
@@ -263,6 +263,65 @@ shutdown:
   timeout: 30s
 ```
 
+### Runtime configuration composition
+
+Dependency ranges remain `ns.dependency` registry entries and exact portable
+versions remain in `wippy.lock`. For local development, a runtime profile can
+replace a locked module with a checkout without changing the lock:
+
+Pass `--config` more than once to compose any set of runtime configuration
+files. Files use the same schema and merge from left to right, so later files
+override matching leaves while preserving unrelated settings:
+
+```sh
+wippy run \
+  --config .wippy.yaml \
+  --config .wippy.dev.yaml \
+  --config config/postgres-history.yaml \
+  --config .wippy.workspace.yaml \
+  --profile workspace
+```
+
+Every explicitly named file must exist. With no `--config`, `.wippy.yaml`
+remains the optional default. The first file defines the project directory used
+to resolve relative runtime paths. Profiles are applied after all files merge,
+in requested order, and CLI overrides such as `--set` are applied last.
+
+Configuration filenames have no reserved meaning. Keep private or
+machine-specific files out of version control using the repository's normal
+ignore policy. For example, the final file above could contain:
+
+```yaml
+version: "1.0"
+
+profiles:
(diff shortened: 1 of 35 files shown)


wolfy-j (approved):
Reviewed: configuration composition, dependency/history behavior, publishing isolation, cross-platform tests, and integration with #471 are ready.

wolfy-j (approved):

PR #465, opened 2026-07-13
fix: refresh static filesystems after registry updates

wolfy-j:
Makes http.static resolve its filesystem from the registry per request, so fs.register replacement events from fs.directory/fs.embed immediately refresh an existing mount instead of leaving it bound to a stale or closed filesystem.\n\nRegression: publishes old and new filesystems through the real fs registry event bus; the same mounted /app/keeper/app.html route must change from old bundle to new bundle.\n\nVerified: go test ./service/http; make build-wippy-local; local Kickside smoke GET /app/keeper/app.html -> 200.


wolfy-j:

PR #468, opened 2026-07-13
fix: preserve declared root dependency ranges

wolfy-j:
Preserves existing root dependency declarations during changeset expansion instead of replacing them with installed lock versions. Regression test covers an unrelated changeset preserving a declared range. Focused test: go test ./boot/deps/hub. Local binary built but not installed.


skhaz (approved):

wolfy-j:

PR #469, opened 2026-07-13
fix: preserve embedded packs across dependency installs

wolfy-j:
### Summary
- limit obsolete embedded-pack cleanup to modules controlled by the active dependency operation
- preserve the root application and unrelated module filesystems during live installs
- add a file-backed regression test covering an application pack while installing CRM

### Root cause
Dependency expansion resolved only the active root graph, but embedded-pack cleanup treated every installed module absent from that partial graph as removed. The registry entries and HTTP mounts remained live while their backing pack handles were closed, producing HTTP 200 responses with truncated or empty bodies.

### Verification
- `go test ./boot/deps/hub`
- `go test -race ./boot/deps/hub`
- built local Linux binary with `make build-wippy-local`
- started `kickside/kickside@0.1.63` in an empty folder
- installed `spiralscout/crm@0.1.18` and `spiralscout/stat-analysis@0.1.4` live
- verified root app, Keeper, Agents, and Knowledge response bodies remained byte-for-byte unchanged after both installs
- verified CRM and Statistical Analysis UI/API routes and all 14 migrations

### DIFF
--- a/boot/deps/hub/dependency_handler.go
+++ b/boot/deps/hub/dependency_handler.go
@@ -254,7 +254,7 @@ func (h *DependencyHandler) Expand(ctx context.Context, op regapi.Operation, sna
 	}
 
 	var effects []regapi.Effect
-	packEffect, err := h.buildEmbedPackEffect(ctx, resolved, snapshot)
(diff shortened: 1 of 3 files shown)


wolfy-j:

PR #470, opened 2026-07-13
fix(hub): resolve dependencies against the declared range, not the resolved pin

skhaz:
## The bug

keeper declares a **range**:

```yaml
component: wippy/dataflow
version: ">=v0.4.10"      # any dataflow at 0.4.10 or newer
```

Hub picks the newest thing that fits and replies:

```json
{ "org": "wippy", "name": "dataflow", "version": "0.5.2" }
```

The range is not in that reply. Hub answered the question and threw away the question itself — there was no field to carry it.

That breaks the lock. `wippy.lock` pins dataflow at `0.4.31`, and before installing, the resolver asks one thing: *is my locked version still allowed?* But the only thing it has to compare against is what hub sent, so the question it actually asks is:

> is `0.4.31` allowed by `0.5.2`? → **no**

Of course not — `0.5.2` is an exact version, not a range, and nothing satisfies it but itself. So the lock is judged invalid, discarded, and the dependency is force-bumped to latest. The lock never had a chance: it was being tested against the wrong thing.

## The fix

Hub now sends the range as well:

```json
{ "org": "wippy", "name": "dataflow", "version": "0.5.2", "constraint": ">=v0.4.10" }
```

and the resolver compares the lock against the constraint instead of the version. Now it asks the right question:

> is `0.4.31` allowed by `>=v0.4.10`? → **yes**

Lock kept. In `resolver.process` that is one line:

```go
- r.addConstraint(dep.Org+"/"+dep.Name, source, dep.Version)               // "0.5.2"      the answer
+ r.addConstraint(dep.Org+"/"+dep.Name, source, declaredConstraint(dep))   // ">=v0.4.10"  the question
```

It falls back to `dep.Version` when the server does not report a constraint, so an older hub still yields a usable graph.

**With no lock, `>=v0.4.10` still resolves to 0.5.2 — that is correct, and unchanged.** The bug was never that hub picked the wrong version; `>=` genuinely permits 0.5.2. The bug is that the range never reached the client, so the lock could not be checked against it.

## Second change, required by the first

`resolveIntersection` joined all of a module's constraints into one string and reparsed it. The parser caps a constraint at 10 parts — a budget sized for one authored range. While every parent contributed the same *pin*, they deduplicated to one and never hit it. Now that parents contribute their own distinct *ranges*, six parents with two-sided ranges make 12 parts, overflow the cap, and a perfectly satisfiable set gets reported as a **version conflict naming every requester**, with the module not resolving at all.

So each constraint is now matched independently, selecting the highest release satisfying all of them (stable preferred, as before). Without this, preserving ranges would break ordinary graphs.

## Relationship to #468

No file overlap; trial-merged, clean, 306/306 combined.

#468 preserves **root** declarations, which the lock already protects (`dependency_handler.go:527` passes `LockedVersions`). But **transitive** deps get a pin from hub, which no older lock can satisfy — so they bump regardless. That is plausibly why #466 bumped everything and was reverted: the root pinning came off while the transitive path still pinned to latest. This PR fixes that half; the two are complementary.

## Testing

Full suite green (306/306). Lint clean. Mutation testing (gremlins): 0 surviving mutants in changed files.

Seven regression tests, each verified to fail when its fix is reverted:
- `TestResolve_LockedVersionSurvivesDeclaredRange` — the bug (`expected 0.4.31, actual 0.5.2` without the fix)
- `TestResolve_ManyParentsWithCompatibleRanges` — the intersection overflow
- `TestResolve_IncompatibleRangesStillConflict` — genuine conflicts still detected, not masked
- `TestGetManifest_DecodesDeclaredConstraintFromWire` — real Connect round-trip

## Order

Needs the hub change (wippy/hub!44) and the proto field (wippy/proto!2). Deploy hub first; this client degrades to current behavior against an un-upgraded hub.

### DIFF
--- a/api/hub/wippy/api/hub/manifest/v1/message.pb.go
+++ b/api/hub/wippy/api/hub/manifest/v1/message.pb.go
@@ -202,15 +202,19 @@ func (x *ModuleManifest) GetProtected() bool {
 
 // Resolved dependency with download info.
 type ResolvedDependency struct {
-	state         protoimpl.MessageState `protogen:"open.v1"`
-	Org           string                 `protobuf:"bytes,1,opt,name=org,proto3" json:"org,omitempty"`
-	Name          string                 `protobuf:"bytes,2,opt,name=name,proto3" json:"name,omitempty"`
-	Version       string                 `protobuf:"bytes,3,opt,name=version,proto3" json:"version,omitempty"`
-	VersionId     string                 `protobuf:"bytes,4,opt,name=version_id,json=versionId,proto3" json:"version_id,omitempty"`
-	Digest        string                 `protobuf:"bytes,5,opt,name=digest,proto3" json:"digest,omitempty"`
-	SizeBytes     uint64                 `protobuf:"varint,6,opt,name=size_bytes,json=sizeBytes,proto3" json:"size_bytes,omitempty"`
-	Download      *v11.DownloadInfo      `protobuf:"bytes,7,opt,name=download,proto3" json:"download,omitempty"`
-	Protected     bool                   `protobuf:"varint,8,opt,name=protected,proto3" json:"protected,omitempty"`
+	state     protoimpl.MessageState `protogen:"open.v1"`
+	Org       string                 `protobuf:"bytes,1,opt,name=org,proto3" json:"org,omitempty"`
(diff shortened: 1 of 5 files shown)


wolfy-j:
sounds like we solved the same issue :) cool

wolfy-j:

PR #478, opened 2026-07-14
fix(registry): resolve baseline dependencies during boot

wolfy-j:
## Root cause

`LoadState` expanded dependency directives only while replaying history. Source declarations in the v0 baseline were published without expansion, so a dependency absent from `wippy.lock` stayed incomplete until a later unrelated dependency changeset re-expanded the full root set.

## Fix

- expand baseline directives before replaying history or publishing registry state
- fail boot when a declared dependency cannot be resolved instead of accepting a partial root
- make unchanged baseline dependency expansion idempotent when the compatible module is already loaded
- retain normal resolution for missing or incompatible modules

## Proof

- regression test: a v0 baseline dependency creates its module entries during `LoadState`
- failure test: baseline expansion failure leaves no published root
- integration test: Analysis installs during boot; a later CRM install emits no Analysis operation or download
- `go test ./... -count=1`
- `make lint` (`0 issues`)
- combined local binary with the pending runtime and WASM fixes: clean app boot with R declared in source but absent from lock and vendor downloaded and initialized `wippy-langs/r@0.2.6`; authenticated Webscout update completed at registry version 1 without changing or reinitializing R

Review only. This PR has not been merged.

### DIFF
--- a/boot/deps/hub/dependency_boot_test.go
+++ b/boot/deps/hub/dependency_boot_test.go
@@ -1,0 +1,137 @@
+// SPDX-License-Identifier: MPL-2.0
+
+package hub
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"github.com/wippyai/runtime/api/payload"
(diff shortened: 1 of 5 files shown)


wolfy-j:

PR #499, opened 2026-07-17
fix(tests): replace timing-dependent synchronization

skhaz:
Removes three independent timing races from the test suite:

- Extends the registry generation-drain hang guard from 1 second to 10 seconds. Successful delivery still completes immediately; the timer only bounds a genuine hang.
- Replaces the process subscription sampler with a deterministic lifecycle assertion taken before `Process.Close`, proving `unlisten` reclaimed all live state instead of relying on 100-microsecond observations.
- Drains native-process stderr before `Wait` closes `StderrPipe`, matching the `os/exec` pipe contract and preventing buffered output loss.
- Makes the application harness fail immediately when a stage fails, uses the same native build tags as Runtime tests, and skips external SQS by default like the other service-backed suites.

## Validation

- `make lint`
- `go test -race -count=20 ./runtime/lua/modules/process ./service/exec/native`
- `./tests/app/test` — 402 application tests passed
- Existing GitHub Linux, Windows, lint, and CodeQL checks are green.


wolfy-j (approved):
Reviewed the complete test change. The timer is only a hang guard around an asynchronous registry-generation drain; increasing it from 1s to 10s does not change runtime behavior or assertions, and the test still fails deterministically on a real hang.

PR #501, opened 2026-07-18
fix(deps): bind history graphs to deployment baseline

wolfy-j:
## Summary
- bind durable dependency resolutions to a deterministic deployment-baseline digest
- re-resolve the final root graph once when the deployment baseline changes, while retaining stored pins for history-owned overlays
- atomically rebind a registry version only across deployment baselines; reject graph mutation within one baseline
- emit a structured warning before repairing stale/legacy resolution state
- carry repaired resolutions through boot and undo/redo transitions

## Why
Registry history is a declarative overlay on the current deployment. Previously, an exact graph stored under an older deployment could overwrite modules loaded by a newer root lock during restart. This could silently restore old package code even while `wippy.lock` showed the new version.

## Protocol
The effective graph is immutable for `(registry version, deployment baseline)`. A changed baseline permits one atomic re-resolution/rebind. An unchanged baseline remains offline/deterministic and cannot rewrite its graph. Legacy unbound graphs can be upgraded once.

## Verification
- `go test ./...`
- `WIPPY_POSTGRES_HISTORY_TEST_DSN=... go test ./system/registry/history/postgres -count=1` against PostgreSQL 16
- root v1 -> v2 restart with persisted history
- transitive add/remove and shared dependency with history-installed plugin
- injected resolution failure leaves state/history unchanged
- second restart performs exact replay without another repair
- undo/redo preserves the v2 root while moving the user overlay
- real Kickside PostgreSQL app: warning emitted once, Keeper 0.5.63 retained over persisted 0.5.62 graph, 83 migrations clean, dependency list/plan/dry-run all HTTP 200, second restart clean

## Addition: fold duplicate root declarations into recorded references

Stacked on the baseline binding above; motivated by the hub-run component development loop (develop crm/gtm/any package against a running hub app without the monorepo).

### Problem
A workspace package must declare its `ns.dependency` entries to resolve standalone (`wippy update`, publish). Loading that same source into a running application through governance fs-sync previously conflicted: `dependency component "X" is already installed as "app.deps:x"`. One source tree could not serve standalone and in-app development.

### Protocol
- A duplicate root declaration for an already-installed component folds into a **recorded Reference** on the durable resolution instead of conflicting. References are facts of the selection: their constraints join the solve, they are verified against the selected version at declaration time, and they reconcile as part of a strict `Roots ∪ References` declaration set.
- References never control install/uninstall. Deleting a reference removes nothing else; deleting the controlling root promotes the reference in the next resolution.
- Preserved conflicts: parameter disagreement, and a fresh duplicate install of a not-yet-installed component (the "update that dependency instead" planning answer is unchanged).
- On a deployment-baseline change, reference constraints participate in the one-shot re-resolution; incompatibility surfaces there with the standard resolver error.

### Compatibility
- `References` is digest-relevant via `omitempty`: an empty set keeps `Digest`/`InputDigest` and the canonical JSON byte-identical to prior releases, so existing applications produce unchanged graphs and history rows. Referenced graphs are content-distinct (a shared digest previously let the content-addressed `resolution_graphs` store return the reference-free payload for a referenced graph — covered by test).
- `InputDigest` remains roots-only: the declared root-set identity used for baseline refresh decisions is unchanged.
- Downgrade constraint: a runtime older than this change cannot reconcile a state that has folded references present (strict root-set count). Same operating note as the baseline binding: upgrade forward, or remove the reference declarations before downgrading.

### Also included
- `BuildDelta` rejects duplicate entry IDs in either input state up front (`duplicate entry id in target state: {ns, name}`), replacing the mid-apply "entry already exists" rollback whose cause (one source loaded through two paths) was invisible at the failure site.

### Verification
- `go test ./...` (305 pkgs) and `-race` across `boot/deps/hub`, `api/registry`, `system/registry/...`.
- Three hardening rounds in `reference_fold_hardening_test.go` + history/model tests: expand semantics (fold recording, fresh-install conflict, declaration-time constraint failure, promotion-on-delete, order-independent election and digests), durable history (content-distinct referenced graphs, sqlite save/close/reopen round-trip, no within-baseline rebind for reference-only changes), reconcile drift (missing/unrecorded/drifted declarations rejected, wildcard normalization, stored-selection satisfaction).
- Live: hub-run `kickside/kickside@0.1.81` + workspace package with its own `ns.dependency` — governance sync folds and records (`references:[...]` in the sqlite graph), restart restores v1 and reconciles the union strictly; keeper MCP loop (sync → edit → re-sync → in-process tests) exercised end to end.

### Hardening (four adversarial rounds)
Election is operation-independent (parameter carrier controls, ties by canonical key, fresh declarations never control — enforced across whole batches); reconciliation folds leniently so committed state can never wedge boot on parameter drift; legacy graphs upgrade via refresh regardless of declaration ID order; reference constraints normalize at the fold. Verified by dedicated regression tests for each of the reviewed failure modes (state poisoning, batch gate bypass, order-dependent legacy upgrade, padded/bare constraints, ID-spelling duplicates) plus full-suite and race runs, and a live hub-run app cycle: sync with folded dependency -> restart -> strict Roots∪References replay.

### Review round: durability and version transitions (326d75d13)
A two-reviewer adversarial pass (invariants → attacks → executed repros) over the registry/history/deps state machine closed six findings:
- **Boot baseline validation**: `LoadState` now rejects duplicate entry IDs in the raw baseline before map construction (previously the last spelling silently won and the fold ran against a truncated declaration set).
- **Semantically unsound graphs are unstorable**: `Valid()` requires a selected module for every declaration and rejects selections a declaration provably excludes under the semver grammar; all four history backends refuse such graphs at the write boundary. `boot/deps/hub/semver` moved to `api/semver` (two import sites) so the model can interpret the constraint spellings it stores; channel pins and exact literals stay fail-open in the model and fail-closed in the hub sweep.
- **Offline legacy upgrade**: a baseline-unbound graph whose stored selection already satisfies every folded reference upgrades in place — partitioned by its recorded roots, no hub round trip — so a cold offline restart cannot wedge on declarations that change nothing about the selection. Unsatisfied references still go through the real resolution.
- **Version-transition baseline identity**: graphs bind the deployment digest of the state they describe (expand binds the post-op snapshot; reconcile compares the transition target). Undo/redo across a change to a baseline-owned declaration now replays both graphs verbatim; previously every crossing re-resolved online and rebound the graph in both directions.
- **Pinned as specification**: adding parameters to a folded declaration transfers control to the carrier (lifecycle safety is promotion-based, not identity-based); committed parameter disagreement boots leniently and surfaces on the next dependency operation as a conflict naming both declarations.
- **Coverage added**: golden pre-change digest generated from the pre-fold code at 7d61cc743; full boot cycle through the real planner + sqlite history including ApplyVersion undo/redo across referenced and promoted versions; referenced round-trips for memory/nil/postgres; `ExpandChanges` delete+create and same-ID delete/recreate batches; controller parameter drift reconciliation.


skhaz (approved):

skhaz (approved):

wolfy-j:
Added for today’s release in `369447e6`:

- raw `wippy pack` snapshots now collect the same manifest-authorized runtime metadata as published application packs;
- named profiles therefore survive `pack -> run --profile` without a local `.wippy.yaml`;
- the existing publisher filter remains authoritative, excluding workspace configuration, unreferenced environment-backed variables, and other machine-local state;
- packs without a `wippy.yaml` remain supported.

Verification:
- `go test ./... -count=1`
- raw Kickside pack booted from an empty directory using only packed `local + sqlite + bootstrap_admin` profiles, loaded 2,731 entries, and served the login page on the selected isolated port.

PR #502, opened 2026-07-19
fix(update): preserve replaced modules' lock rows across lock rebuilds

wolfy-j:
wippy update rebuilds the lock from hub resolution only; tracked replacements are excluded from resolution by design, so the replaced module's own row (commonly a test harness or local workspace module with root: true) was silently dropped on every update. The replacement then pointed at a module the graph no longer declared and every ns.requirement of the harness failed to link ("no dependency parameter found").

Fix: carry the old lock's module rows forward for tracked replacements on both the full and targeted update paths. Rows the new resolution did produce stay authoritative; a rowless replacement is not invented.

Verified: unit tests for the carry (root row kept, resolved row wins, no invention); end to end against a standalone module harness — wippy update now preserves the acme/starter-harness root row and the harness suite passes afterwards. Full ./cmd/wippy/cmd and ./boot/deps/lock suites green.

### DIFF
--- a/cmd/wippy/cmd/update.go
+++ b/cmd/wippy/cmd/update.go
@@ -205,6 +205,7 @@ func runUpdate(cmd *cobra.Command, args []string) error {
 	// Preserve all replacements from old lock file
 	if oldLockObj != nil {
 		preserveReplacements(newLockObj, oldLockObj.GetTrackedReplacements())
+		preserveReplacedModuleRows(newLockObj, oldLockObj)
 	}
 
 	// Save lock file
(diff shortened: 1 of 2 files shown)


skhaz (approved):

wolfy-j:
Release review: I do not recommend merging this implementation. It carries a stale module row forward from the old lock specifically for deprecated lock-file replacements. That is compatibility/compensation state, and the path is narrower than the canonical workspace replacement model (GetTrackedReplacements, not the effective workspace overlay). The structural fix is to remove the deprecated tracked-replacement resolution branch or make replacement roots explicit inputs to lock reconstruction; copying an old version/hash/root row is not an authoritative rebuild.

wolfy-j:
Closed without replacement. This preserves stale module rows only for deprecated replacements persisted in wippy.lock. The canonical workspace.replacements overlay already keeps replacement configuration outside the lock while retaining Hub-resolved module rows, so it does not need this carry-forward compensation. Harnesses must use the workspace configuration.

wolfy-j:
Closed without replacement. This preserves stale module rows only for deprecated replacements persisted in wippy.lock. The canonical workspace.replacements overlay already keeps replacement configuration outside the lock while retaining Hub-resolved module rows, so it does not need this carry-forward compensation. Harnesses must use the workspace configuration.

PR #503, opened 2026-07-19
fix(deps): adopt active application roots on live update

wolfy-j:
## Problem

Installing an updated root application through Hub fails when its published artifact redeclares deployment-root dependencies. Those declarations are intentionally unowned in the boot baseline, so the generic ownership conflict check rejects the application that originally authored them.

## Change

- pass the active deployment-root module set from the lock into operation planning
- allow only an active root application to adopt an unowned `ns.dependency` entry that its own desired artifact declares
- keep conflicts for ordinary host entries, non-root packages, other module owners, and declarations that remain dependency roots
- apply the same rule during normal expansion and resolution reconciliation

## Compatibility / risk

This does not change resolution, lock format, or general registry ownership. The only changed case is an unowned dependency declaration with the same ID as a declaration in an actively booted root application. A customized unowned declaration with that exact ID will now be replaced by the application update; unrelated unowned entries remain protected.

## Verification

- `go test ./...`
- PostgreSQL Kickside production-tag boot restored 64 roots
- real Hub UI updated `kickside/kickside` from 0.1.89 to 0.1.95 without process restart
- running PID remained unchanged and System UI returned HTTP 200

### DIFF
--- a/boot/deps/hub/dependency_handler.go
+++ b/boot/deps/hub/dependency_handler.go
@@ -343,9 +343,10 @@ func (h *DependencyHandler) expand(
 	}
 
 	additional, err := (operationPlanner{resolver: h.resolver}).plan(snapshot, combined, operationPlanOptions{
-		originalKey:       idKey(op.Entry.ID),
-		controlledModules: controlledModules,
-		mutableModules:    mutableModules,
+		originalKey:           idKey(op.Entry.ID),
+		controlledModules:     controlledModules,
+		mutableModules:        mutableModules,
+		deploymentRootModules: h.deploymentRootModules(),
 	})
 	if err != nil {
 		return regapi.DirectiveResult{}, err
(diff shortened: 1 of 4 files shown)


skhaz (approved):

wolfy-j:
Release review blocker: this PR must not be stacked with #504. I tested #502 + #503 + #504 + #505 on current main; TestDependencyHandler_Expand_ActiveApplicationAdoptsDeploymentRootDependencies fails because #504 correctly keeps the root application's declared namespace dependency as DependencyRoot=true, while entryConflictForPlan here only permits adoption when desired.DependencyRoot is false. More importantly, cold boot now tags published root entries with module provenance and root provenance; allowing an active app to claim an unowned declaration is repair/compensation logic for malformed old topology. #504 restores live/cold topology parity at ingestion and should supersede this takeover exception.

wolfy-j:
Closed as superseded by #504. The takeover exception permits a root application to adopt an unowned registry/user overlay declaration and conflicts with the canonical DependencyRoot topology. #504 fixes cold/live ingestion parity without changing ownership authority.

PR #504, opened 2026-07-19
fix(deps): preserve root application dependency topology

wolfy-j:
## Summary

- preserve explicit deployment-root and user-overlay dependency authority across application updates, rollback, and restart
- compose every version transition from the deployment baseline plus the selected history resolution
- bind dependency parameters through each package's declared `ns.definition` root instead of deriving namespaces from package names
- treat a fully-qualified `namespace:name` parameter as an exact registry requirement address
- keep unpacked/local module loading inside its source boundary by excluding nested `.wippy` runtime state

## Semantic model

- `component` identifies an artifact (`org/module`); it is not a registry namespace.
- The artifact's single root `ns.definition` declares the namespace used for bare requirement parameters.
- Fully-qualified requirement IDs are exact addresses and can intentionally wire another package's public requirement.
- `DependencyRoot` and the persisted resolution graph determine deployment ownership. No namespace, including `app.deps`, is hard-coded as a root.
- Root application entries, user-authored overlay roots, and transitive module entries retain separate provenance.
- Packed modules, unpacked modules, local replacements, live updates, and cold restart use the same source and linking rules.

## Root causes fixed

The live Hub loader previously reloaded root-application dependency declarations as module-owned transitives, stripping root authority. Historical transitions could then compose against stale intermediate graphs instead of the current deployment baseline.

The linker also inferred registry namespaces from package spelling. That breaks valid packages such as a plural artifact exporting a singular namespace and cannot support deliberately unrelated package/namespace names.

Finally, a local replacement rooted at a workspace directory could recursively ingest `.wippy/vendor` as if those dependencies belonged to the replacement module.

## Verification

- `go test ./... -count=1`
- `make lint`
- exact candidate binary: `nightly-20260719-75810d1-9-g8ce6cae21`
- PostgreSQL Kickside history restored at head v7 with 42 roots and 44 resolved modules; HTTP returned 200 after cold restart
- real unpacked `userspace/users` package resolved through its declared `userspace.user:definition` namespace
- focused coverage includes root application update, downgrade/undo/redo, restart replay, user overlays, arbitrary namespaces, exact cross-namespace parameters, packed/unpacked provenance, and nested `.wippy` exclusion

This supersedes #503. Its application-takeover exception is neither needed nor permitted by this model; #502 and #503 remain closed.

### DIFF
--- a/boot/build/stages/errors.go
+++ b/boot/build/stages/errors.go
@@ -68,12 +68,6 @@ func NewDecodeDependencyError(entryID string, cause error) apierror.Error {
 	return apierror.New(apierror.Internal, fmt.Sprintf("failed to decode dependency %s", entryID)).WithCause(cause)
 }
 
-func NewInvalidDependencyComponentError(entryID, component string, cause error) apierror.Error {
-	return apierror.New(apierror.Invalid, fmt.Sprintf("invalid dependency component %s", component)).
-		WithDetails(attrs.NewBagFrom(map[string]any{"entry_id": entryID, "component": component})).
-		WithCause(cause)
-}
-
 func NewRequirementError(requirementName, namespace string, cause error) apierror.Error {
 	return apierror.New(apierror.Internal, fmt.Sprintf("requirement %s in namespace %s failed", requirementName, namespace)).WithCause(cause)
 }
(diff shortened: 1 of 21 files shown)


skhaz (approved):

wolfy-j:
Release review: the ingestion-level fix is the correct direction. It derives root status from the selected root module in the lock, not from an app.deps namespace convention, and makes cold/live loading produce the same topology. Focused CI and the PostgreSQL lifecycle proof are strong. Merge this without #503; the two are semantically incompatible as currently written.

PR #507, opened 2026-07-19
fix(run): make Hub deployments restart offline

wolfy-j:
## Summary

- materialize every verified Hub-run pack under the project-local vendor directory recorded by `wippy.lock`
- run the first launch from those exact local artifacts instead of the disposable global cache paths
- load published runtime defaults and named profiles from the lock-selected root pack during bare `wippy run`

## Root cause

`wippy run org/module@version` wrote `wippy.lock` while leaving downloaded packs only in the global cache. Bare `wippy run` later loaded the lock vendor directory, where those artifacts did not exist. It also resolved configuration without the selected root pack metadata, so named profiles from a published application failed before entry loading.

## Contract

A Hub reference bootstraps a local deployment. Once the lock is written, that deployment restarts from the exact locked project-local artifacts without Hub resolution or download. The single lock-selected root pack is the only package allowed to contribute published host runtime defaults; dependency packs cannot configure the host.

There is no version scan, namespace convention, or cache fallback.

## Verification

- `go test ./cmd/wippy/cmd -count=1`
- unit coverage for exact artifact materialization and lock-root profile loading
- PostgreSQL Hub boot → live root package update → bare offline restart lifecycle (in progress on the stacked release candidate)

### DIFF
--- a/cmd/wippy/cmd/run.go
+++ b/cmd/wippy/cmd/run.go
@@ -177,7 +177,15 @@ func runWithUseCase(cmd *cobra.Command, args []string, useCase string) error {
 
 	logger.Info("initializing runtime", zap.String("memory_limit", formatBytes(memLimit)))
 
-	cfg, err := loadRuntimeConfig(cmd, logger)
+	// A Hub deployment is restarted from wippy.lock, without resolving the Hub
+	// reference again. Its selected root pack is therefore the same authority
+	// for published runtime defaults as it was on the first run.
+	runtimeDefaults, err := loadLockRootRuntimeDefaults(defaultLockFile, logger)
+	if err != nil {
+		logger.Error("failed to load deployment runtime defaults", zap.Error(err))
+		return err
+	}
+	cfg, err := loadRuntimeConfigWithDefaults(cmd, logger, runtimeDefaults)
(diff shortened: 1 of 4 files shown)


wolfy-j:
PostgreSQL lifecycle proof completed on a production-tag local binary stacked with #504, #506, and #499:

1. Fresh empty folder + fresh PostgreSQL 16 database.
2. `wippy run kickside/kickside@0.1.97` resolved 44 modules and launched from 44 project-local vendor packs.
3. Keeper live-installed `kickside/kickside@0.1.98`; HTTP stayed available.
4. Clean shutdown.
5. Restarted with bare `wippy run` and `WIPPY_CACHE_DIR` pointing to an empty directory.
6. No Hub resolution/download occurred; empty cache remained at 0 files.
7. Registry restored PostgreSQL history version 1, System UI returned 200, and inventory plus root-owned entries reported `kickside/kickside@0.1.98`.
8. `wippy.lock` remained the immutable deployment baseline at 0.1.97, while both baseline and overlay packs remained locally available.

This verifies the intended baseline + history-overlay model across a real restart.

skhaz (approved):

PR #522, opened 2026-07-27
feat(http): support method-agnostic endpoints

skhaz:
## Summary

- support explicit quoted `method: "*"` on `http.endpoint`
- register method-agnostic endpoints as methodless `net/http.ServeMux` patterns so dynamic SPA fallbacks coexist with static submounts
- skip synthetic `OPTIONS` registration for all-method endpoints
- convert `ServeMux` registration panics into structured route-conflict errors
- expose `{path...}` rest wildcard values through route parameters

Jira: https://jira.spiralscout.com/browse/EE2-2329

## Verification

- [x] `TMPDIR=/private/tmp/wippy-tests make test`
- [x] `TMPDIR=/private/tmp/wippy-tests make lint`
- [x] `go test ./api/service/http ./service/http -race -count=1`
- [x] independent adversarial review completed with no remaining findings

## Performance

- method-agnostic and concrete catch-all routing both remain at 80 B/op and 5 allocs/op
- 100-route rebuild median improved from 457.6 µs/op to 178.2 µs/op
- rebuild allocations improved from 3,589 to 2,988 allocs/op

### DIFF
--- a/api/service/http/config.go
+++ b/api/service/http/config.go
@@ -27,6 +27,9 @@ const (
 	// Static identifies a static file server component
 	Static registry.Kind = "http.static"
 
+	// MethodAny registers an endpoint for every HTTP method.
+	MethodAny = "*"
+
 	// ServerID is the key used to identify the server in configuration metadata
 	ServerID string = "server"
 	// RouterID is the key used to identify the router in configuration metadata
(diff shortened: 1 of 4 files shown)


wolfy-j:
Do we still have req:method() right?

skhaz:
Yes. `req:method()` is unchanged and still returns the actual incoming HTTP method (for example, `GET`, `POST`, or `PROPFIND`). `method: "*"` only makes route matching method-agnostic; it does not alter the request method exposed to Lua.

wolfy-j:
great

PR #552, opened 2026-08-03
feat(lua): support process entry security

skhaz:
## Summary
- add `security` configuration to source and bytecode Lua process entries
- propagate entry policy metadata through factory registration and apply it before process execution
- preserve inherited security baseline across upgrades while removing superseded entry grants
- add `SOURCE_DATE_EPOCH` support and `publish --dry-run --output` for reproducible prospective publish artifacts
- fix the existing Windows stderr test to use `cmd` rather than PowerShell, which held the inherited pipe open

## Validation
- `make test`
- `make lint`
- focused race-enabled process, security, host, actor, publish, and native executor tests
- Windows native executor package cross-compilation
- targeted source-date and publish-output mutants killed
- built the CLI and ran the full QuickJS validation, including source and packed consumer smoke, against it
- independent security and release reviews: approved

## Performance
`BenchmarkSchedulerSubmit` on Apple M4, GOMAXPROCS=1, 3s x 5: median improved from 3093 ns/op at base to 2283 ns/op.

### DIFF
--- a/api/process/process.go
+++ b/api/process/process.go
@@ -14,6 +14,7 @@ import (
 	"github.com/wippyai/runtime/api/registry"
 	"github.com/wippyai/runtime/api/relay"
 	"github.com/wippyai/runtime/api/runtime"
+	"github.com/wippyai/runtime/api/security"
 )
 
 // System identifies the process system in the event bus.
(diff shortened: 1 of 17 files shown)


wolfy-j:
Closing as superseded. Process-entry security landed in #555, canonical security resolution landed in #558, and the Windows stderr fix landed in #567. The remaining upgrade-security replacement semantics and reproducible publish-output work should be handled as separate focused changes against current main; this branch's frame-local security state and non-atomic publish output should not be carried forward.

PR #606, opened 2026-08-26
fix(deps): defer workspace replacement loading

skhaz:
## Why

A workspace may declare replacements for modules that are not selected by the current lock graph. Treating every replacement as a runtime load path makes an unrelated or temporarily absent workspace directory break boot and lets unselected source participate in pack, install, and artifact preparation.

## What

- Make the existing `GetModuleLoadPaths` contract return only app source and modules selected by the lock graph.
- Apply a replacement only when its source module is selected; selected replacements retain version, digest, root, and source-root semantics.
- Keep dependency discovery separate: `wippy update` scans effective workspace replacement roots directly so an unselected local module can still contribute dependency constraints before resolution.
- Keep `GetLoadPaths` as the path-only view of the same canonical selection; no boot-only API or parallel load-path vocabulary.

The unrelated offline dependency-chain diagnostics were removed from this PR.

## Verification

- `go test -race -count=1 ./boot/deps/lock ./cmd/internal/entries ./cmd/wippy/cmd`
- `golangci-lint` on the changed dependency and CLI packages
- Targeted coverage for unselected replacement exclusion, selected root fallback, selected replacement metadata, and workspace replacement dependency scanning

### DIFF
--- a/boot/deps/lock/lock.go
+++ b/boot/deps/lock/lock.go
@@ -417,7 +417,8 @@ type ModuleLoadPath struct {
 // App source has empty Module/Version.
 // Replacement paths carry Module from replacement "from" and retain the
 // selected lock version when one exists. The replacement changes source
-// ownership, not the resolved release identity.
+// ownership, not the resolved release identity. Replacements for modules that
+// are not selected by the lock graph are resolution inputs, not load paths.
 func (l *Lock) GetModuleLoadPaths() []ModuleLoadPath {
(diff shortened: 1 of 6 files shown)


wolfy-j (approved):
Verified replacement ownership boundary: boot/install load only selected graph sources; update retains all declared replacement inputs for resolution; malformed declarations remain rejected. Focused lock and command tests pass.

PR #607, opened 2026-08-26
fix(registry): migrate persisted operation ownership state

wolfy-j:
## Problem

Registry ownership and dependency-root state must be owned by the registry, not inferred from author `meta`. Existing persisted rows need a one-time conversion before normal replay. An established deployment also could not restart on a new host from its history database because its baseline module graph existed only in `wippy.lock`.

## Fix

- Store owner/root in `Entry.Registry`; leave author `meta` and `data` unchanged.
- Run one registry-history child migration (`1.0` to `1.1`) before replay. It derives owner/root from the active WAPP baseline and the stored dependency resolution, then rewrites persisted entry operations through the existing SQLite/PostgreSQL migration ledger.
- Expose one atomic registry snapshot to Lua with entry owner/root and the selected module graph. There is no provenance side map or metadata fallback.
- Persist the deployment baseline inside the existing dependency-resolution checkpoint: root module plus exact module version and artifact digest.
- On a lockless restart, materialize those exact artifacts through the Hub, populate the existing source registry, and use the same module loader, normalization pipeline, and embedded-WAPP registration as lock-backed boot.
- Preserve that baseline through dynamic install, update, uninstall, rollback, restart, and stale-lock/local-replacement startup handling.

There is no new history table, public migration API, operation mode, alternate resolver, or second entry-loading path.

## Rollout

An existing deployment starts once with its current lock so the baseline is attached atomically to the current resolution. Later starts may use the history database without a lock or pre-existing WAPP cache. A pre-upgrade database started without either the persisted baseline or its lock fails closed.

## Proof

- Restored the current C PostgreSQL dump into an isolated database.
- Upgrade boot loaded the 2,905-entry deployment baseline, replayed head v397, reached `runtime ready`, and stored `kickside/kickside` plus 47 exact baseline artifacts without re-resolving the graph.
- A newly created empty directory with no `wippy.lock` and no `.wippy` cache fetched the recorded artifacts, loaded the same 2,905 entries through the canonical loader, replayed v397, registered embedded filesystems, completed all bootloaders, and reached `runtime ready`.
- A second lockless boot reached `runtime ready` again; the head resolution digest was unchanged and registry/application migrations performed zero writes.

## Validation

- `go test ./...`
- `make lint` — 0 issues
- SQLite, PostgreSQL, memory-history, dependency lifecycle, loader, and registry tests
- `git diff --check`


skhaz (approved):

skhaz (approved):

skhaz (approved):

wolfy-j:
Commit 3 — `fix(registry): keep one resident identity across retained entries`.

Second production defect after the boot fix: every keeper REST call answered 500 `registry snapshot state failed: wippy/dataflow: ... module has conflicting provenance records`. Root cause (reproduced in sim on the production history + app tree): the deployment lock pins `wippy/dataflow 0.7.5` while the registry root resolves `0.7.14`; `userspace.dataflow:execution_identity` was removed in 0.7.14 but is retained by `operationPlanner.plan` because `app:dataflow_identity_binding` still names it — and it stayed stamped with the 0.7.5 artifact identity, so one module carried two resident identities and `ProvenancedState.Validate()` refused the snapshot. Not latent in history (legacy fold yields 0 module-owned records); created at boot by baseline + resolution reconcile.

Fix in `boot/deps/hub`: `residentIdentities` folds the single artifact identity a module is resident at; `adoptResidentIdentity` moves a retained entry's record onto it (entry-scoped `Root` preserved); the planner emits a provenance-delta `EntryUpdate` for retained entries whose module moved. A module with no resident artifact left keeps its record. Strict invariants untouched. TDD: `boot/deps/hub/retained_entry_provenance_test.go` fails with the exact production string before the fix.

Proof over the production history: boot 0 errors, `GET /keeper/hub/dependencies` 200 (64 modules), `POST .../plan` 200 with a complete install payload; repeat boot converges (no re-emitted delta); pure-legacy and fresh-install conditions still pass; full boot provenance dump 70 modules / 5092 entries / zero conflicts; `./system/registry/... ./boot/deps/... ./api/registry/...` green on sqlite and postgres.

Known, left visible: `preserveImmutableResidentEntry` compares `Version` only, never `Digest` — an immutable module resident at the same version but a different digest can fork identity the same way; not reachable on this data.

wolfy-j:
Correction to the previous comment: the retain-and-re-stamp fix ("keep one resident identity across retained entries") is **withdrawn** in favor of cde05d78ad `fix(deps): reject partial module identity updates`, which refuses the fork at its source instead of re-stamping a retained entry with an artifact identity its content does not come from. The root-cause analysis stands: on the affected instance the deployment lock pins `wippy/dataflow 0.7.5` while the registry root resolves `0.7.14`, and `app:dataflow_identity_binding` still names `userspace.dataflow:execution_identity`, which 0.7.14 removed. The operator-side alignment for that instance is being verified in sim against the current branch HEAD; results follow. The regression test from the withdrawn commit (production shape: retained entry under a moved module) is worth carrying as an assertion of the new refusal — will propose it once the verdict is in.

skhaz (approved):

skhaz (approved):

PR #628, opened 2026-09-02
fix(eventbus): make unsubscribe a terminal delivery barrier

wolfy-j:
## Problem

Event bus shutdown could panic with `send on closed channel`. A canceled context allowed `Unsubscribe` to return without removing the subscription, and an unsubscribe arriving after `Stop` marked the bus closed was acknowledged while the dispatcher could still hold an in-flight send reference. Callers could then close their delivery channel before the sole sender was fenced.

The hardening review also found a second ownership bug: `Subscribe` could enqueue a request, return `context.Canceled` without an ID, and then have the dispatcher install the unreachable subscription. Three direct cluster consumers could also stop reading while their live subscription remained backpressured, preventing the dispatcher from ever reaching their unsubscribe barrier.

## Design

- Define `Unsubscribe` as a non-cancelable ownership barrier.
- During bus draining, wait for dispatcher exit before acknowledging unsubscribe.
- Make repeated and concurrent `Stop` calls wait for the same terminal barrier.
- Once a subscribe request is queued, wait for the dispatcher's serialized ownership decision; reject a canceled request before installing it and return no ID on error.
- Give the Raft bootstrap watcher, Raft membership handler, and global name-registry listener owned subscription contexts. Cancel before they stop consuming or enter the unsubscribe barrier, including bootstrap watcher's natural-completion path.
- Document the actual send boundary: delivery is ordered/lossless while publisher and subscriber contexts stay active; cancellation may abort queued or in-progress delivery.

This uses the existing single-dispatcher model. It adds no recovery, panic suppression, queue, VM, or delivery machinery.

## Critical-delivery audit

- Registry transactions and entry application, Lua/WASM process/function registration, queue declarations, and store resource handover prepare correlated `AwaitService` waiters before publishing. Cancellation or missing replies fail the operation instead of reporting silent success.
- A 128-reply correlated burst test exceeds the shared subscriber channel buffer and proves every distinct reply is routed without loss.
- Cluster membership/bootstrap/global-name event consumers are advisory: reconciliation, polling, Raft state, and anti-entropy repair missed work. Their shutdown contexts are now owned explicitly so backlog cannot deadlock shutdown.
- `eventbus.NewSubscriber`, KV watcher, topology listener, and process event dispatcher already cancel their owned subscription contexts before waiting at `Unsubscribe`.

## Regression proofs

- queued canceled subscribe is rejected and retains neither ID nor channel
- unsubscribe preserves 256 preceding events exactly once and in order, then delivers none afterward
- stop preserves 128 preceding events for each of two subscribers exactly once and in order, then delivers none afterward
- critical `AwaitService` burst routes all 128 correlated replies
- Raft bootstrap (explicit stop and natural completion), membership handler, and global name registry cancel before the terminal barrier

Targeted barrier, critical-delivery, and consumer lifecycle tests passed 100 iterations under `-race`.

## Validation

- `go test -race ./...`
- targeted regression tests: `-race -count=100`
- rebased onto current `main` (`github.com/wippyai/go-lua v1.5.19`)
- `git diff --check`

## Performance

Five-run local comparison against the original PR head:

- `BenchmarkSend`: ~97 ns median vs ~102 ns baseline, 0 allocs/op
- `BenchmarkSendParallel`: within run-to-run noise, 0 allocs/op
- `BenchmarkSendMultipleSubscribers`: within run-to-run noise, 0 allocs/op
- `BenchmarkSubscribeUnsubscribe`: ~1.11 us median vs ~1.06 us baseline, unchanged 10 allocs/op

There is no measurable hot-path regression; the new context check is on subscribe processing, not send delivery.

### DIFF
--- a/api/event/event.go
+++ b/api/event/event.go
@@ -31,12 +31,16 @@ type (
 	// Bus defines the functionality of an event bus.
 	Bus interface {
 		// Subscribe subscribes a channel to events from a specific system.
+		// The caller must not close the channel until Unsubscribe returns;
+		// that return is the barrier proving the bus can no longer send to it.
 		Subscribe(context.Context, System, chan<- Event) (SubscriberID, error)
 
-		// SubscribeP subscribes a channel to events matching a specific system and kind.
+		// SubscribeP has the same channel ownership and Unsubscribe barrier as Subscribe.
 		SubscribeP(context.Context, System, Kind, chan<- Event) (SubscriberID, error)
 
-		// Unsubscribe removes a subscription using its SubscriberID.
+		// Unsubscribe removes a subscription using its SubscriberID. It is a
+		// non-cancelable safety barrier: when it returns, no in-flight or future
+		// bus operation can send to that subscription.
 		Unsubscribe(context.Context, SubscriberID)
 
 		// Send publishes an event to the bus.
--- a/api/event/event.go
+++ b/api/event/event.go
@@ -33,6 +33,9 @@ type (
 		// Subscribe subscribes a channel to events from a specific system.
 		// The caller must not close the channel until Unsubscribe returns;
 		// that return is the barrier proving the bus can no longer send to it.
+		// Once the request is accepted for dispatch, Subscribe waits for the
+		// ownership decision even if the context is canceled. On error, the bus
+		// does not retain the channel.
 		Subscribe(context.Context, System, chan<- Event) (SubscriberID, error)
 
 		// SubscribeP has the same channel ownership and Unsubscribe barrier as Subscribe.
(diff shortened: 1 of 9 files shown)


skhaz (approved):

wolfy-j:
@skhaz please re-review the rebased hardening commit. It adds deterministic proofs for queued-subscribe ownership, exact no-loss/in-order delivery before unsubscribe and stop barriers, a 128-reply AwaitService burst, and cancellation-before-barrier fixes for Raft bootstrap/membership and global name registry. Full `go test -race ./...` is green locally; hot send benchmarks remain 0 allocs with no measurable regression.

skhaz (approved):

PR #629, opened 2026-09-03
Support process.send from function handlers

skhaz:
# Reason for This PR

A `function.lua` handler can call `process.send` through `funcs.call`. The runtime used a `go-lua` version that lost the sole resumable Go frame for a tail-position system yield. Recovered Go panics also returned no Go stack and wrote no diagnostic log. This change also adds regression coverage for https://github.com/wippyai/go-lua/issues/37.

This pull request depends on https://github.com/wippyai/go-lua/pull/41. It uses the exact public commit while that pull request waits for approval.

## Description of Changes

- update `go-lua` to the commit that contains the yield fix and recovered-panic diagnostics
- log the same recovered Go panic stack that the caller receives
- add runtime tests for a tail-position external yield and recovered panic logging
- add application tests for statement, protected-call register-pressure, and tail-position `process.send` calls through `funcs.call`

## Validation

- `make lint`: zero issues
- `make test` with race detection and documented external-service skips: passed
- `make build-wippy-local`: passed
- built runtime end-user smoke with `app.test.funcs:process_send_runner`: passed
- `go mod verify`: passed
- independent code review: no remaining findings

The full application harness stopped before test execution because Docker is not available on this host. The focused application smoke test uses the same application entries and passed.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -66,7 +66,7 @@ require (
 	github.com/tree-sitter/tree-sitter-php v0.24.2
 	github.com/tree-sitter/tree-sitter-python v0.25.0
 	github.com/tree-sitter/tree-sitter-typescript v0.23.2
-	github.com/wippyai/go-lua v1.5.18
+	github.com/wippyai/go-lua v1.5.20-0.20260903130651-4d15e597625b
 	github.com/wippyai/module-registry-proto-go v0.0.1
 	github.com/wippyai/tree-sitter-markdown v0.0.3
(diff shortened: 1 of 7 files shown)


wolfy-j:
Superseded by #630, which moved runtime main to the released go-lua v1.5.19 containing the tail-yield fix. This PR also depends on an unreleased panic-callback API that is unrelated to process.send and should not land. Any useful W1 regression coverage can be preserved separately without the pseudo-version or panic machinery.

PR #639, opened 2026-09-05
fix: preserve WebSocket results, defer HTTP status, and correct TTY startup

wolfy-j:
The documentation sweep in wippyai/docs#35 exposed dropped WebSocket operation results, premature HTTP header commitment, an omitted HTTP client option type, and TTY startup ordering that could lose initial events.

- Return `true, nil` or `false, error` from WebSocket send/ping through the yield result handler.
- Defer the selected HTTP status until the first body write/flush or handler completion. `set_status(201); set_header(...); write_json(...)` now sends the chosen status and headers together; status-only responses such as 204 remain supported.
- Include `overlay_network` in the exported HTTP client request options type.
- Subscribe to events before starting input in the TTY demos. Align the toolkit's sample events and layout width with the current types so both demo apps lint cleanly.

Validation: affected HTTP, HTTP client, WebSocket and TTY Go suites pass, including race checks. Regression checks fail against the previous implementations. Built the runtime, linted both TTY demo apps, and verified rendering and normal exit of tty-demo, norton and tty-proof through pseudo-terminals.

The HTTP change stores one pending status integer per request and keeps streaming writes. No performance benchmark was run.

### DIFF
--- a/api/service/http/context.go
+++ b/api/service/http/context.go
@@ -56,6 +56,7 @@ type RequestContext struct {
 	r               *http.Request
 	w               http.ResponseWriter
 	responseHandled bool
+	responseStatus  int
 }
 
 // GetRequestContext retrieves HTTP request context from FrameContext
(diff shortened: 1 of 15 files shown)


wolfy-j:
Split into independent PRs: #640 (WebSocket results), #641 (HTTP status/header ordering), #642 (HTTP overlay type declaration), and #643 (TTY startup ordering). Each replacement preserves the corresponding fix and regression coverage; local affected checks and full lint pass.

PR #670, opened 2026-09-08
feat(raft): add bounded cancellable proposal waiting

wolfy-j:
Add optional `raft.ContextApplier` / `Node.ApplyContext` so consumers can stop waiting for a Raft proposal without pretending to cancel a commit. This is a prerequisite for the context-aware KV/naming integration tracked in #664.

The implementation reserves count and command-byte capacity before copying caller input. An admitted proposal retains both reservations until its underlying Apply finishes, including after caller cancellation. Capacity exhaustion waits on release notifications with cancellation/shutdown support; oversized commands are rejected before allocation or count-slot waiting. Lifecycle checks run after contention and immediately before invoking Apply. No retry or rollback is implied by an uncertain result.

Boot exposes positive integer budgets `cluster.raft.max_pending_applies` (128 by default) and `cluster.raft.max_pending_apply_bytes` (64 MiB). These cover only unresolved ApplyContext work and its owned command payloads. Persistent log/FSM storage and existing direct Apply calls are outside those limits. The existing Apply timeout argument retains its queue-admission meaning; the context bounds caller waiting.

Validation: real HashiCorp Raft future cancellation, caller-buffer ownership, retained count/byte capacity after cancellation, recovery after actual completion, oversized rejection, shutdown while waiting for bytes, and a reproduced cancellation-versus-admission-lock regression. Standalone race suites pass for Raft (3.714s), boot and real-Raft cluster integration (50.355s); final boot checks and scoped lint pass.

This PR provides the bounded primitive. KV/naming adoption is separate and is not claimed to be completed here. No merge requested.

### DIFF
--- a/api/cluster/raft/config.go
+++ b/api/cluster/raft/config.go
@@ -4,6 +4,12 @@ package raft
 
 import "time"
 
+// DefaultMaxPendingApplies bounds context-aware future tracking per node.
+const DefaultMaxPendingApplies = 128
+
+// DefaultMaxPendingApplyBytes bounds retained command copies for unresolved work.
+const DefaultMaxPendingApplyBytes = 64 << 20
+
 // Config holds configuration for a Raft node.
 //
 // Raft defaults to a diskless control plane (in-memory stores): cluster state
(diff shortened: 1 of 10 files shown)


skhaz (approved):

skhaz (approved):

wolfy-j:
Closing for now: there is no current caller of ApplyContext or ContextApplier, including the minimized naming work. This would otherwise add an exported interface, public Raft configuration, two boot knobs, and a new admission subsystem without a consumer. We can reintroduce the smallest required primitive when KV or naming actually adopts cancellable proposal waiting.

PR #690, opened 2026-09-08
fix(metrics): exclude in-flight sends before collector shutdown

wolfy-j:
## Problem

A producer could pass the collector's atomic `closed` check immediately before shutdown closed `recordCh`. The later send then raced with channel close and could panic. This surfaced during teardown of the race-enabled cluster harness and is reproduced by the focused regression test against the pre-fix source.

## Fix

- Add a send-admission barrier around the final closed check and non-blocking enqueue.
- Make shutdown set `closed` only after excluding admitted senders, then signal the export loop to drain without closing the producer-owned record channel.
- Make concurrent `Close` calls wait for the same completed shutdown.
- Reject exporter registration after shutdown begins while preserving exporters whose registration won the race.
- Close every registered exporter once and join their errors.

No exporter method runs while the send-admission lock is held, and the change adds no worker or timer.

## Regression coverage

- Concurrent producers versus shutdown: every attempt is exported or counted as dropped; no send/close race or panic.
- Post-close records are counted as dropped.
- Concurrent `Close` callers wait for exporter shutdown and receive the same result.
- Exporter close errors are joined and retained across repeated `Close` calls.
- Registration racing with shutdown is linearizable: the exporter is either accepted and closed exactly once, or rejected with `metrics collector is closed`.

The focused race regression immediately reports a channel send/close race and `send on closed channel` panic on the pre-fix commit. On this head, `go test -race -count=10 ./service/metrics/...`, 50 repeated focused shutdown runs, and `go vet ./service/metrics/...` pass. The full CI matrix is green: lint, Ubuntu, Windows, native application, PostgreSQL/SQLite CDC race suites, and CodeQL.

## Cost

The admission lock adds measurable synchronization cost to the deliberately saturated zero-allocation microbenchmarks (roughly 2x in local runs) while retaining zero allocations. This is the explicit cost of making shutdown linearizable with in-flight producers.

### DIFF
--- a/service/metrics/collector.go
+++ b/service/metrics/collector.go
@@ -19,6 +19,7 @@ type collector struct {
 	wg        sync.WaitGroup
 	dropped   atomic.Uint64
 	exportMu  sync.RWMutex
+	recordMu  sync.RWMutex // excludes in-flight send admission from shutdown
 	closed    atomic.Bool
 }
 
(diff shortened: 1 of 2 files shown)


skhaz (approved):

skhaz (approved):

wolfy-j:
@skhaz The shutdown hardening changed the head after the earlier approval. Please re-review current commit 538ec943: concurrent Close now waits, late exporter registration is rejected/linearized, exporter errors are joined, and the receiver no longer closes the producer-owned channel. All checks are green.

PR #694, opened 2026-09-09
feat(exec): process groups for native children, group signalling and proc:pid()

wolfy-j:
## Gap

The native executor starts pipe-backed children in the runtime's own process group and session, and `Signal`/`Stop` address the direct pid only. Measured against a child `sh -c 'echo $$; sleep 300 & echo $!; wait'`:

| Case | Direct child | Grandchild |
|---|---|---|
| `close()` (TERM) | dies | survives, reparented |
| `close(true)` (KILL) | dies | survives |
| child ignoring TERM, `close()` | dies after the grace | survives |
| owning Lua process exits without `close` | cleaned up | survives |

A harness such as Claude Code or Codex spawns tool subprocesses and MCP servers as grandchildren, so whole-tree cleanup is impossible through the runtime API today.

## Change

- `ProcessOptions.ProcessGroup` (per exec, `process_group = true` from Lua) and `NativeExecutorConfig.ProcessGroup` (executor default, per-command override in both directions).
- Pipe-backed children start with `Setpgid`; the PTY path keeps `setsid` (already a group leader).
- `Signal`, `Stop`, `close()` with its grace escalation, and the owner-exit cleanup path all deliver to the group when enabled.
- `proc:pid()` (new `ProcessIdentity` capability) so callers can record execution evidence.
- `exec.run` policy metadata carries `process_group`.
- Windows: the option returns `errors.UNAVAILABLE` instead of being ignored.
- Docs in `runtime/lua/modules/exec/spec.md`.

Group signalling controls what remains in the created group; a descendant that calls `setsid` escapes it, which a container executor confines by construction.

## Tests

Tests were written first and fail on main. `service/exec/native/pgroup_test.go` and `runtime/lua/modules/exec/pgroup_test.go` drive real trees through the Go API and the Lua bindings, including the with/without differential pair, TERM-ignoring children, the PTY path, executor default versus override, and the runtime resource-store cleanup on owner exit.

`go test ./service/exec/... ./runtime/lua/modules/exec/... ./api/service/exec/...`, `go vet`, golangci-lint and a `GOOS=windows` build all pass.

### DIFF
--- a/api/service/exec/api.go
+++ b/api/service/exec/api.go
@@ -20,9 +20,13 @@ const (
 
 // ProcessOptions defines options for creating a new process
 type ProcessOptions struct {
-	Env     map[string]string
-	PTY     *PTYOptions
-	WorkDir string
+	Env map[string]string
+	PTY *PTYOptions
+	// ProcessGroup places the child in its own process group so that signals
+	// addressed to the process reach its descendants as well. Nil selects the
+	// executor default.
+	ProcessGroup *bool
+	WorkDir      string
 }
 
 type PTYOptions struct {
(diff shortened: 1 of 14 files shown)


skhaz (approved):

skhaz (approved):

wolfy-j:
Superseded by #696. The process-group implementation commit is patch-equivalent and fully preserved there, alongside the combined acceptance coverage and platform fixes. Closing this duplicate keeps one canonical review/merge path.

PR #695, opened 2026-09-09
feat(exec): report process exit on a channel without consuming the handle

wolfy-j:
## The gap

`proc:wait()` sets `closed = true` and drops the handle *before* it yields. Once
any coroutine calls `wait()`, every other method on that process -- `write_stdin`,
`signal`, `close`, the streams -- reports `process is closed`.

A supervisor that has to keep feeding a child and steering it with signals while
also learning when it exits could do neither: it either consumed the handle with
`wait()` or never learned the exit. Reading stdout to EOF is not a substitute
either -- a grandchild inherits the pipe and can hold it open long after the
child itself is gone. That consumption was also barely documented ("Automatically
closes the process").

## The behavior

**`proc:done()` -> channel, error.** A one-shot channel usable in
`channel.select` / `:receive()`. It delivers exactly one value when the child
exits:

| Field | Type | Notes |
|-------|------|-------|
| `code` | integer | exit code, or `128 + signal` for a child killed by a signal |
| `signal` | integer | the signal that killed the child; absent when it exited on its own |
| `error` | error | set only when the exit could not be observed at all |

The handle stays open: `signal()`, `write_stdin()`, the streams and `close()`
keep working until `close()` is called. Calling `done()` twice returns the same
channel; the channel closes behind its single value, so a second `receive()`
reports `nil, false` rather than blocking.

**One reap, owned by the runtime.** The wait moves into the process wrapper:
`done()`, `wait()` and `close()` all funnel through it, so the child is waited on
exactly once whichever gets there first, and every caller observes the same exit.
The wait yield now carries a wrapper that defers to that reap, so `wait()` after
`done()` has fired returns the recorded exit code instead of failing a second
`cmd.Wait` ("exec: Wait was already called"). `wait()` keeps its consume-the-
handle semantics otherwise, now spelled out in `spec.md`. Runtime-owned cleanup
(resource store close -> `close(false)`) still reaps correctly when `done()` was
used, and does not wait twice.

**Exit semantics on signal death.** Classification moves to
`api/service/exec.ClassifyExit`, one place instead of an inline branch in the
dispatcher. Go reports no exit code for a signaled child, so the code is
normalized to the conventional `128 + signal` the native executor already
produces, and the signal number is carried beside it. Executors that report only
a code (a container exit) report no signal. A non-zero exit stays a code, never
an error; `error` is reserved for failing to observe the exit at all.

`types.go` gains `exec.ProcessExit` / `exec.ProcessExitChannel` and the `done`
method; `spec.md` documents `done()`, the exit value, and what `wait()` consumes.

## Tests

`runtime/lua/modules/exec/done_test.go` drives real children through the Lua
bindings on a real engine process (inline pool, exec + stream dispatchers, relay
node routing the exit package back):

- a coroutine selects on `done()` while the main flow writes stdin, reads the
  echo back, then signals -- exit arrives as `code 143, signal 15`
- `wait()` after `done()` returns 7 for a child that exited 7 (fails with
  "exec: Wait was already called" without the shared reap)
- a grandchild holds stdout open for 5s while the direct child exits 3:
  `done()` fires immediately, with a separate test proving the premise (the pipe
  really has no EOF after the child is reaped)
- the child is waited on exactly once across `done()` and `close()`

`api/service/exec/exit_test.go` covers classification, including a really
signal-killed child.

`go test ./runtime/lua/modules/exec/... ./service/exec/...`, `go vet` and
golangci-lint are clean. (`runtime/lua/modules/tty` has one failure,
`TestSurfacePresentClearsRemovedRows`, that is pre-existing on other branches
from main and unrelated.)

## Relationship to #694

Independent of PR 694 (process groups). Both touch
`runtime/lua/modules/exec/process.go`: 694 changes how a child is signaled,
this changes how its exit is observed. Whichever lands second needs a small
merge in `close()` / the process struct, no design conflict.

### DIFF
--- a/api/service/exec/exit.go
+++ b/api/service/exec/exit.go
@@ -1,0 +1,82 @@
+// SPDX-License-Identifier: MPL-2.0
+
+package exec
+
+import (
+	"errors"
+	osexec "os/exec"
+	"syscall"
+)
+
+// ExitStatus describes how a child process finished.
+//
+// Code is the exit code the child reported. A child killed by a signal has no
+// exit code of its own, so it is reported as 128+Signal, the encoding shells
+// use and the one the native executor produces.
+//
+// Signal carries the signal that killed the child, or 0 when the child exited
+// on its own or the executor reports only a code, as a container exit does.
+//
+// Err is set only when the exit could not be observed at all: a wait that could
+// not be performed, a transport failure to a remote executor. A non-zero exit
+// is not an error, it is Code.
+type ExitStatus struct {
+	Err    error
+	Code   int
+	Signal int
+}
+
+// ExitCoder is an error carrying the exit code of the process it describes.
+type ExitCoder interface {
+	ExitCode() int
+}
+
+// ExitReporter is a Process that owns the reap of its child and can therefore
+// report the outcome more than once. Wait can only be performed once, so a
+// caller that has an ExitReporter must use it instead: several parts of a
+// supervisor may need the exit of the same child.
+type ExitReporter interface {
+	AwaitExit() ExitStatus
+}
+
+// ClassifyExit turns the error Process.Wait returns into an exit status.
+func ClassifyExit(err error) ExitStatus {
+	if err == nil {
+		return ExitStatus{}
+	}
+
+	var coder ExitCoder
+	if !errors.As(err, &coder) {
+		return ExitStatus{Err: err}
+	}
+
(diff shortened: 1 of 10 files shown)


skhaz (approved):

wolfy-j:
Superseded by #696. The done-channel implementation commit is patch-equivalent and fully preserved there, alongside the combined acceptance coverage and platform fixes. Closing this duplicate keeps one canonical review/merge path.

PR #703, opened 2026-09-09
feat(app): isolate explicit client launch from owner execution

wolfy-j:
## Purpose

Let a Builder-produced executable select an application-specific client or owner invocation without changing ordinary Wippy runtime semantics.

## Design

- All implementation and contracts live in `cmd/app`, following #727. No `api/application`, public statelock package, Lua API, or boot-component discovery is added.
- `Options.Launch` is an explicit native callback. It can run a client directly or call the one-shot, scoped owner runner. The application owns discovery, authentication, admission and attachment errors.
- The owner runner alone acquires application-state ownership, prepares native owner resources, opens deployment/history, executes the existing runtime command, and cleans up before unlocking.
- Actual contention returns `app.ErrBusy`; filesystem errors are not contention. No automatic attachment, retry, or owner fallback occurs.
- Recovery, update and runtime tooling bypass the callback. Explicit state selection and registry-history ownership remain fixed.
- An escaped/repeated owner runner is rejected. Returning while it is active cancels and joins it and reports a lifecycle error.

#726's embedded-baseline selection is a separate application option, not a launch result. Builder and Bee need corresponding explicit wiring; the former implicit `PrepareLaunch` component integration is intentionally removed. Do not merge until those integration checks are complete.

## Validation

- `go test -race ./cmd/app -count=10` passes.
- Pinned golangci-lint on `./cmd/app/...`: zero issues.
- Windows amd64 non-CGO application test binary cross-compiles (not executed on Windows).
- Tests cover client-only state isolation, busy-owner routing without store access, lock I/O errors, preparation/seed failure cleanup under ownership, reserved operation bypass, expired owner credentials, fixed registry history, and runner lifetime enforcement.
- No changes outside `cmd/app`; no docs or README files added to the repository.

PR #726, opened 2026-09-10
feat(app): select embedded baseline through executable options

wolfy-j:
## Purpose

Let Builder-produced applications explicitly choose whether ordinary startup follows the activated deployment or the executable's embedded bundle, without changing ordinary Wippy runtime semantics.

## Design

- `cmd/app.Options.Baseline` selects `activated` (also the empty/default value) or `embedded`; invalid values fail before launch.
- Embedded ordinary startup uses the digest-qualified bundle and retains the selected state's registry history.
- Explicit base recovery retains separate registry history. Update and runtime tooling continue selecting the activated deployment and bypass the application launch callback.
- The policy is executable configuration, independent of #703's launch callback. No `LaunchPlan` flag, core API package, README or public locking package is added.
- Stacked on the revised #703. All changes are under `cmd/app`.

## Validation

- Ten race runs of `./cmd/app` pass.
- Tests cover digest changes, unchanged activated deployment, shared ordinary history paths, isolated recovery history, stale activation, and non-destructive seed failures.
- Builder-generated old/new binary upgrade and authored-history/replacement acceptance remain required before merge; path-selection tests alone do not establish those guarantees.

This work is separate from the focused v0.3.41a release candidate.

### DIFF
--- a/cmd/app/baseline.go
+++ b/cmd/app/baseline.go
@@ -1,0 +1,38 @@
+// SPDX-License-Identifier: MPL-2.0
+
+package app
+
+import (
+	"fmt"
+	"path/filepath"
+)
+
+// embeddedDeployment returns the immutable, digest-scoped deployment for a
+// bundled application. Ordinary embedded startup and explicit base recovery
+// share these files, but Run gives them different registry history paths.
+func embeddedDeployment(state string, bundle Bundle) string {
+	return filepath.Join(state, "base", bundleID(bundle))
+}
+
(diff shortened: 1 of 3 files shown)


skhaz (approved):

skhaz (approved):

wolfy-j:
Superseded by #766 (stacked on #765): same selection, now tested end to end through Run, with the duplicated predicate and the loose assertions removed.

PR #738, opened 2026-09-12
test(http): give listener probe requests a context

wolfy-j:
The port-zero listener tests added in #737 call `http.Client.Get`, which fails the repository's `noctx` lint rule. Build requests with the test context and send them through the existing bounded client.

Production behavior is unchanged. Focused port-zero, named-host and probe-failure tests pass.

PR #739, opened 2026-09-12
feat(exec): authorize per-process Docker bind mounts

wolfy-j:
Processes using one Docker executor need distinct project and retained-home mounts. Add typed per-process bind mounts to Go `ProcessOptions` and Lua `executor:exec` options, with a separate `exec.mount` check on each source before process creation. Native execution rejects mount options.

Validate absolute sources and container targets, reject duplicate request/configured bind targets, and copy options before retaining them. Docker keeps its existing static volume syntax; collision checks include Windows host sources targeting Unix container paths. The source permission is a lexical path check, not symlink confinement, and refers to the Docker daemon's host.

Validation: race tests pass for `api/service/exec`, all `service/exec` packages and `runtime/lua/modules/exec`. Two live Alpine containers verify independent mounted homes, terminal input and resize. Lua tests cover typed options and denial before factory invocation, and native tests cover unsupported mounts. Scoped golangci-lint reports zero issues. The exec API suite also passes on actual Windows after fixing Unix daemon-path validation for Windows clients.

CI on the preceding revision passed Linux and native Lua. The Windows failure is corrected in the latest commit. Full CI lint currently reports three unchanged HTTP test `noctx` violations, addressed separately by #738.

This provides the runtime primitive. Managed harness Docker configuration and recovery remain Bee integration work.

## Follow-up commits
- Static mount targets are validated once, at executor construction; `NewProcess` checks only what varies per process. Redundant slice copies after `Clone` removed.
- Denials are `PERMISSION_DENIED` (were `INVALID`) for `exec.run`, `exec.mount` and executor access; a refused mount names `source`/`target`/`read_only` in details. Bee's `render_configuration.lua:48` asserts the old kind and changes to `"PermissionDenied"` when this lands.
- `Mount.Validate` reports through apierror constructors with `ErrInvalidMount`/`ErrDuplicateMountTarget` as the cause; every `ErrorIs` assertion still holds.
- Coverage: policy is evaluated for every mount with its target and mode (`TestExecutorExecMountPolicyScopesEachSource`, `…SeesAttributes`); one refused source refuses the whole process; a path that fails validation never reaches policy (`TestExecutorExecValidatesMountsBeforePolicy`).
- Spec has a `mounts` usage example. Native executors keep refusing mounts with `ErrMountsUnsupported`; a native confinement option is a separate, queued feature.

### DIFF
--- a/api/service/exec/api.go
+++ b/api/service/exec/api.go
@@ -4,7 +4,11 @@
 package exec
 
 import (
+	"fmt"
 	"io"
+	"path"
+	"path/filepath"
+	"strings"
 
 	"github.com/wippyai/runtime/api/registry"
 )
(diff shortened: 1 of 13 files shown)


skhaz (dismissed):

skhaz (approved):

skhaz (approved):

skhaz (approved):

PR #740, opened 2026-09-13
feat(cli): select terminal hosts from command declarations

wolfy-j:
Installing a dependency that contains a second `terminal.host` currently prevents an application's named command from starting: host auto-detection becomes ambiguous even though the application knows which host it uses.

Allow `meta.command.host: namespace:name` to declare that host. An explicit `--host` retains precedence; commands without a declaration retain single-host discovery. Malformed declarations, missing hosts and non-terminal hosts fail before process creation. Selecting a host does not add actor or policy grants.

Validation: focused metadata/security/host tests and the full `cmd/wippy/cmd` race suite pass; current golangci-lint reports zero issues. Tests cover two installed hosts, undeclared ambiguity, missing/wrong-kind targets and malformed metadata. Bee's actual offline restart reproduction reached this ambiguity after its separately diagnosed artifact-cache failure was removed; final Bee integration acceptance follows the runtime build.

### DIFF
--- a/cmd/wippy/cmd/command_host_test.go
+++ b/cmd/wippy/cmd/command_host_test.go
@@ -1,0 +1,73 @@
+// SPDX-License-Identifier: MPL-2.0
+
+package cmd
+
+import (
+	"errors"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	ctxapi "github.com/wippyai/runtime/api/context"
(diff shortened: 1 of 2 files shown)


skhaz (dismissed):

skhaz (approved):

skhaz (approved):

skhaz (approved):

PR #741, opened 2026-09-13
fix(boot): honor offline dependency restoration

wolfy-j:
Registry startup currently overrides dependency access with `Online`, so restarting an installed application may unexpectedly contact the Hub. A missing local artifact can stall boot or fail when the network is unavailable.

Core startup now requests verified offline restoration, and `PrepareRestore` honors that policy for both rooted deployments and source checkouts. Explicit install/update operations keep their existing network behavior. Recorded module identities and registry history are unchanged; this does not add cache locations or substitute newer artifacts.

Validation: the full dependency-handler and core-component race suites pass. The restore tests prove zero Hub queries/downloads on an offline cache miss, successful explicit online materialization, and subsequent offline cache reuse. Stable cache retention across embedded executable updates is a separate application-runner change.

### DIFF
--- a/boot/components/core/registry.go
+++ b/boot/components/core/registry.go
@@ -124,7 +124,10 @@ func Registry() boot.Component {
 			if err != nil {
 				logger.Warn("dependency handler disabled", zap.Error(err))
 			} else if depHandler != nil {
-				if err := depHandler.PrepareRestore(ctx, hist); err != nil {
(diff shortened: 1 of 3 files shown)


skhaz (dismissed):

skhaz (approved):

skhaz (approved):

skhaz (approved):

skhaz (approved):

PR #742, opened 2026-09-13
fix(app): preserve dependency artifacts across embedded updates

wolfy-j:
An embedded executable update selects a new deployment vendor directory while registry history still refers to previously installed module artifacts. Startup consequently attempts to download artifacts that already exist in older local deployments.

Keep dependency artifacts in an application-state cache using the existing `registry.dependency_vendor_dir` setting. Seed embedded packs and import retained exact immutable artifacts, including installed modules absent from deployment locks. Verify bytes while copying, publish atomically without replacing existing content, and preserve registry history, deployment locks and module digests.

This PR stacks on #726 because `cmd/app` is not on main. Offline-only registry restore is independently reviewed in #741; explicit command host selection is #740. None needs to be merged to review this change.

Validation: app suite and focused race tests pass, including changed bundles retaining installed artifacts, symlink rejection, concurrent identical publication, read failure and digest mismatch. Bee's combined executable offline acceptance is in progress and is not claimed here.

### DIFF
--- a/cmd/app/artifact_cache.go
+++ b/cmd/app/artifact_cache.go
@@ -1,0 +1,406 @@
+// SPDX-License-Identifier: MPL-2.0
+
+package app
+
+import (
+	"bytes"
+	"crypto/sha256"
+	"encoding/hex"
+	"errors"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/wippyai/runtime/boot/deps/graph"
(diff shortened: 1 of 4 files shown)


skhaz (dismissed):

skhaz (approved):

wolfy-j:
Superseded by #760: same goal, seeded through the hub's own cache contract instead of a second copy of it, with the silent lock-row and active.json failures surfaced.

PR #743, opened 2026-09-13
feat(exec): expose optional process identity on terminal sessions

wolfy-j:
Managed terminal callers lose access to the optional host process identity when `attach_terminal()` transfers ownership from `exec.Process`. Add `TerminalSession:pid()` so a caller can record that identity for later reconciliation.

Attachment remains asynchronous. PID reads return a retryable unavailable error until startup completes, without acquiring the process startup lock; unsupported identity and failed startup remain explicit errors. The Lua result is `integer?` plus an optional error. A PID or terminal completion does not prove that a process group is gone.

Stacked on #726 (`feat/embedded-default-deployment-pr`) to match the terminal attachment API used by Bee. No remote-monitor or mesh changes.

Validation: affected exec and terminal proxy race suites pass; Go vet passes; repository-pinned golangci-lint v2.13.2 reports 0 issues. Tests cover exclusive PTY ownership, optional identity, delayed startup, failure, unsupported identity and typed guarded PID use. Bee's pinned toolchain builds with the composed patch; native window integration acceptance is still in progress.

### DIFF
--- a/runtime/lua/modules/exec/module_test.go
+++ b/runtime/lua/modules/exec/module_test.go
@@ -627,6 +627,13 @@ type mockPTYProcess struct{ mockProcess }
 
 func (*mockPTYProcess) Resize(int, int) error { return nil }
 
+type mockPTYIdentityProcess struct {
+	*mockPTYProcess
+	pid int
+}
+
+func (p *mockPTYIdentityProcess) Pid() (int, error) { return p.pid, nil }
+
 func TestTakePTYProcessTransfersExclusiveOwnership(t *testing.T) {
 	l := lua.NewState()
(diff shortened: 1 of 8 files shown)


skhaz (dismissed):

skhaz (approved):

wolfy-j:
Closing: `process:pid()` already exposes the child pid (`runtime/lua/modules/exec/process.go`); a second pid accessor on the terminal session duplicates it. The `!bridge.Started()` gate also masks that `native.ProcessExecutor.Pid()` blocks on the startup lock — if that blocking matters, the fix is lock granularity in `native`, not a gate in the Lua layer.

PR #744, opened 2026-09-13
feat(fs): add atomic file publication capability

wolfy-j:
A retained application configuration currently has to be truncated in place or removed before recreation through Lua. A failed write can leave partial content, and resolving a changed parent path can redirect a later write. Add `fs:writefile_atomic(path, content)` through an optional `fs.AtomicWriteFS` capability.

The directory provider on Linux/macOS pins verified parent directory handles, refuses symlink parents and nonregular targets, writes an exclusive temporary file, syncs/closes it, and renames within the pinned directory. Directory-sync failure after rename reports that publication happened but durability is uncertain. Unsupported providers/platforms refuse explicitly. This does not add compare-and-swap or change existing write methods.

Lua accepts strings up to 8 MiB and requests mode 0600. Provider failures use fixed error messages so file content cannot leak through diagnostics. Existing filesystem grants and backend permissions still govern access.

Validation: filesystem API, directory and Lua module race suites; repository-pinned lint for changed packages; Darwin arm64 and Windows amd64 directory-package cross-compilation. Tests exercise real Lua/backend publication, unsupported/denied operations, input bounds, path/link refusal, parent replacement into a sibling, concurrent whole-file readers/writers, temporary cleanup and write/sync/close/rename/post-publication failures. macOS is cross-compiled, not executed; Windows implementation is explicitly unsupported.

This is a runtime capability only. Bee's retained configuration consumer and real-provider cold recovery are separate follow-up acceptance. No runtime merge or Bee runtime-pin change is requested here.

### DIFF
--- a/api/fs/errors.go
+++ b/api/fs/errors.go
@@ -21,3 +21,8 @@ var (
 	ErrNotEmpty    = errors.New("directory is not empty")
 	ErrBusy        = errors.New("filesystem resource is busy")
 )
+
+var (
+	ErrAtomicWriteUnsupported = errors.New("atomic file publication is unsupported")
+	ErrPublishedSyncFailed    = errors.New("file published but directory sync failed")
+)
--- a/api/fs/fs.go
+++ b/api/fs/fs.go
@@ -82,3 +82,13 @@ type (
 		GetFS(name string) (FS, bool)
 	}
 )
+
+// AtomicWriteFS is an optional whole-file publication capability. Implementations
+// must publish either all content or none, without following symbolic-link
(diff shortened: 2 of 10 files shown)


skhaz (dismissed):

skhaz (approved):

skhaz (approved):

wolfy-j:
Superseded by #761: the same atomic publication as an option on fs.writefile, with the backend cause preserved, the permission demand aligned with an ordinary write, and the test-only production interfaces removed.

PR #745, opened 2026-09-13
fix(terminal): keep shutdown progressing during blocked input

wolfy-j:
A PTY child that stops reading stdin can block the terminal event loop inside `WriteStdin`. Context cancellation then cannot send SIGTERM; even direct `RequestClose` cannot escalate if the child ignores SIGTERM.

Move the existing shutdown escalation into one joined watcher per running proxy. It observes cancellation and close intent independently of terminal input, preserves the existing grace periods, and closes owned I/O and cancels abandoned waits on timeout. Run retains normal process/output completion; cancellation, signal failures and forced-shutdown errors remain observable. Public exec and terminal APIs are unchanged.

Validation:
- Deterministic regressions fail on current main for missing TERM on cancellation and missing KILL after direct close; fixed tests also cover reaping and signal-error preservation.
- Race suites pass for terminal, native/Docker executors and Lua exec. Final proxy/Lua exec race run includes the opt-in real Docker regression.
- The real Docker test uses an already-local immutable Alpine image, a non-reading container, and a 16 MiB paste; it performs no pull and cleans up only its own process. Run with `go test -race -tags integration ./service/terminal/proxy -run TestNativeDockerInputCancellation`.
- Scoped lint passes with the Makefile's golangci-lint version, including the integration test.

This fixes cancellation and direct close while input is blocked. It does not add Docker container attachment APIs or complete Bee's managed Docker integration. An event-channel close still follows earlier input; callers needing asynchronous shutdown use the existing `RequestClose` method.

### DIFF
--- a/service/terminal/proxy/docker_input_cancellation_integration_test.go
+++ b/service/terminal/proxy/docker_input_cancellation_integration_test.go
@@ -1,0 +1,91 @@
+//go:build integration
+
+// SPDX-License-Identifier: MPL-2.0
+
+package proxy
+
+import (
+	"context"
+	"strings"
+	"syscall"
+	"testing"
+	"time"
+
+	"github.com/moby/moby/client"
+	"github.com/stretchr/testify/require"
+	execapi "github.com/wippyai/runtime/api/service/exec"
+	ttyapi "github.com/wippyai/runtime/api/tty"
+	dockerexec "github.com/wippyai/runtime/service/exec/docker"
+	"go.uber.org/zap"
+)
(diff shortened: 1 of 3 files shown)


skhaz (dismissed):

skhaz (approved):

skhaz (approved):

wolfy-j:
Superseded by #763: same shutdown behavior, with the cancellation cause carried through beginShutdown instead of re-injected at Run's exit.

PR #746, opened 2026-09-14
feat(lua): add bounded TOML subtree insertion

wolfy-j:
Bee needs to compose one admitted TOML subtree into a private provider configuration without exposing a broad Lua table codec or rewriting user/project files.

This adds a deterministic `toml.insert(document, path, source)` module. It accepts only a source containing the selected subtree, refuses existing targets, preserves TOML value types, and bounds input bytes, path size, decoded depth/entries, and output bytes. The boot component registers the module and its type manifest.

Validation:
- `go test -race -count=1 ./runtime/lua/modules/... ./boot/components/runtime/lua/...`
- `go vet ./runtime/lua/modules/toml ./boot/components/runtime/lua`
- `git diff --check`

The installed golangci-lint cannot decode the repository's Go 1.27 export data, so its result is unavailable from this workstation.


skhaz (approved):

wolfy-j:
Superseded by #762: a general `toml` encode/decode codec shaped like json/yaml; composition (the former `insert`) is the caller's decode → set → encode.

PR #747, opened 2026-09-14
feat(app): allow executable-selected default state

wolfy-j:
A standalone application currently falls back only to the OS user configuration directory, so an executable such as Bee cannot select one default state root per project before its launch callback runs.

This adds an optional `cmd/app.Options.DefaultStateDir` resolver. It runs only when `--state-dir` is absent; an explicit flag retains precedence, an empty or failed resolution refuses before launch, and the resolved directory remains fixed in `LaunchRequest`.

Validation:
- `go test -race ./cmd/app -count=10`
- `go vet ./cmd/app`

This PR is stacked on #766 (which replaced #726) and adds no Lua API, second lock, retry path, or Bee-specific policy.

## Restack
#726 was closed in favor of #766, so this branch now carries #766's branch merged in and targets it; it retargets to main once #766 lands. The resolver's failure and its empty selection are typed errors (`NewDefaultStateResolveError`, Internal with the resolver's cause; `ErrEmptyDefaultState`, Invalid) instead of `fmt.Errorf`, and the refusal test asserts the kinds and the preserved cause.


skhaz (approved):

skhaz (approved):

wolfy-j:

PR #755, opened 2026-09-17
fix(registry): stop blaming a missing listener for a slow handler

wolfy-j:
## What

Two small changes, no behaviour change to the deadline itself:

1. `BusRunner` reported a wait that expired as `no listener responded - check if listener is registered for this kind`. The runner cannot distinguish "nobody is listening" from "the listener is still working", and both land here, so the message now names both. I lost time chasing a missing registration that was never missing.
2. The wasm function manager logs how long a module took to load and compile, so this cost is visible instead of being inferred from a gap between log lines.

## Why

An application failed to boot on v0.3.41a/v0.3.42a with:

```
event handler timeout - no listener responded
  {"id": "kickside.convert.doc2md:converter", "kind": "function.wasm", "operation": "entry.create", "timeout": "30s"}
```

The listener was registered and busy the entire time. With the log added here:

```
wasm.func  wasm module compiled  {"path": "doc2md.wasm", "bytes": 3828286, "component": true,
                                  "load": "6.8ms", "compile": "28.75s"}
registry.runner  event handler timeout ...   (one second later)
```

`function.wasm` compiles eagerly and synchronously inside the create handler, and for this component that is ~28s of genuine codegen against a fixed 30s budget. Ruled out by measurement: memory limit (same at 1G/4G/8G), GC (`GOGC=400`), parallelism (`GOMAXPROCS` 1/8/32). Why the compile is that slow is wippyai/wasm-runtime#18: registering host profiles turns on a whole-component asyncify transform that doubles the module (3.82MB → 7.80MB, ~11s) before wazero compiles twice as much code.

## What this does not fix

Raising `registry.event_wait_timeout` only moves the guess, since compile time scales with the module. The real fix is a handler that can signal it accepted the work and is still running (so the wait keeps its purpose of catching a missing or wedged listener), or compiling outside the registry operation with the pool waiting at first call. That is a protocol decision for this repo, so it is not attempted here.

Related: the same deadline used to fire for `function.lua` when one type check was slow — fixed at the source in wippyai/go-lua#46, where a non-converging flow solve made a 1348-line library take 26.9s.


wolfy-j:
The Lint job fails on `service/http/server_startup_test.go` (`noctx`: `(*net/http.Client).Get` at lines 63, 134, 171). That file is not touched by this PR and the same calls are on `main` today, from #737 — so any PR against `main` inherits it. Rebased onto current `main`; everything else is green.

skhaz (approved):

PR #760, opened 2026-09-18
fix(app): seed the dependency cache through the hub's cache contract

wolfy-j:
Supersedes #742. Same goal — embedded application updates keep already-downloaded dependency artifacts instead of re-fetching at startup — without a second copy of the cache contract.

**Stack:** this branch is #742's branch (which carries #726) merged with #757's `fix/dependency-access-capability`, because the redo touches `boot/deps/hub/module_artifacts.go`, a file #757 introduced. Review the last commit; the rest is #726/#742/#757.

## What changed
- **One owner of the immutable cache layout.** `boot/deps/hub` exports three symbols: `ImmutableWappRelativePath`, `ImmutableWappDigest` (its inverse — also replaces hub's own hand-rolled filename parsing in `soleImmutableArtifact`), and `PublishImmutableArtifact` (containment + skip-if-present + private staging + verify + link, built on the existing `publishVerifiedArtifact`). `cmd/app/artifact_cache.go` drops `immutableRelativePath`, `immutableFilenameDigest`, `sha256Digest`, `publishBytes`, `publishArtifact`, `copyArtifact`, `verifyRootArtifact`, `verifyCachedPath`, `createRootTemp`, `pathWithin`.
- **Malformed lock rows no longer vanish.** An unparsable module name or a malformed digest fails the seed with a structured error (`TestSeedDependencyCacheReportsUnparsableRetainedModuleName`, `...MalformedRetainedModuleDigest`). A row with no hash pins no content and is skipped, which is correct for a content-addressed cache (`TestSeedDependencyCacheAcceptsRetainedModuleWithoutDigest`). The embedded-pack loop had the same latent `name, _ := ParseName(...)`; both go through one helper now.
- **Corrupt `active.json` surfaces** instead of being swallowed by `err == nil` (`TestSeedDependencyCacheReportsCorruptActivation`). This was the only path reading it during embedded startup.
- `bundle.validate()` before the seed is removed: `Bundle.Seed` validates first, and the seed now verifies every byte against the digest it stores under. Drops a per-launch re-hash of every embedded pack.
- Errors in the reworked file go through `cmd/app/errors.go` apierror constructors.

The three publication contract tests moved with the contract to `boot/deps/hub/artifact_cache_test.go` (concurrent identical publishers, reader failure leaves nothing, digest mismatch leaves nothing) plus a private-file-leak assertion and a layout round-trip test.

## Verification
`go build ./...`; `go test` on `cmd/app`, `boot/deps`, `boot/components`; `-race -short` on `cmd/app` and `boot/deps/hub`; lint 0 issues; gofmt clean. The offline-restore test asserting zero Hub calls passes unchanged.

## Noted, not changed
Cache org subdirectories are created `0o755` by hub instead of `0o700` by cmd/app; the cache root stays `0o700`, so nothing is newly reachable. The rest of `cmd/app` still uses `fmt.Errorf`; converting the package is a separate change.

### DIFF
--- a/boot/components/core/registry.go
+++ b/boot/components/core/registry.go
@@ -124,7 +124,10 @@ func Registry() boot.Component {
 			if err != nil {
 				logger.Warn("dependency handler disabled", zap.Error(err))
 			} else if depHandler != nil {
-				if err := depHandler.PrepareRestore(ctx, hist); err != nil {
+				// Startup restores installed artifacts locally. Downloads belong
+				// to explicit install/update operations, never implicit recovery.
+				restoreCtx := regapi.WithDependencyAccess(ctx, regapi.DependencyAccessVerifiedOffline)
+				if err := depHandler.PrepareRestore(restoreCtx, hist); err != nil {
 					return nil, fmt.Errorf("prepare dependency restore: %w", err)
 				}
 				registryOpts = append(registryOpts,
--- a/boot/deps/hub/dependency_boot_test.go
+++ b/boot/deps/hub/dependency_boot_test.go
@@ -42,8 +42,10 @@ func TestDependencyHandlerPrepareRestoreMaterializesRecordedArtifacts(t *testing
 	sum := sha256.Sum256(artifact)
 	digest := "sha256:" + hex.EncodeToString(sum[:])
 	downloads := 0
+	requests := 0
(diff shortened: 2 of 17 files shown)


skhaz (approved):

wolfy-j:

PR #764, opened 2026-09-18
chore(naming): gofmt strong.go

wolfy-j:
`system/topology/namereg/global/strong.go` landed unformatted in e258b11257; `gofmt -l` on main is not empty until this is in. Formatting only, no code change.


wolfy-j:
Closing: this applies Go 1.27 gofmt's layout for a multi-value return of composite literals, which golangci-lint's bundled gofmt rejects (Lint fails on this branch at strong.go:171 under the same `make lint` CI runs). main's current formatting is the one that passes the project gate. The formatter disagreement goes away when golangci-lint ships a printer aligned with Go 1.27; nothing to change in the tree until then.

PR #766, opened 2026-09-18
feat(app): select embedded baseline through executable options

wolfy-j:
Supersedes #726. Stacked on #765 (base is `feat/application-launch-admission-clean`); retarget to `main` once #765 merges.

## What the PR is for
An embedded application binary can start from the deployment its own bundle digest names instead of the activated one, chosen by the executable's `Baseline` option. Runner commands keep addressing the activated deployment.

## Cleanup over #726
- **The option is now tested end to end** through `Run`, not just the pure selector: `TestRunSelectsDeploymentForBaseline` (10 subtests: `embedded`, `activated`, default, `embedded` + runner command → activated, `embedded` + `--base` → embedded; each direct and through `Launch`) and `TestRunRefusesUnknownBaselineBeforeLaunch` (invalid value refused before launch, callback not invoked, no state dir). Mutation-checked: removing the runner-command guard, forcing the embedded flag, or deleting the validation each fails the corresponding test.
- The duplicated `ordinary` predicate is gone; the call site uses #765's single `reservedCommand` and names the local for what it decides (`embedded`).
- `baseline.go` comment states what the function does instead of what it does not.
- `TestEmbeddedBaselineSelectionIgnoresStaleActivation` asserts the exact deployment path instead of `NotEmpty`.
- The two tests that mostly exercised `Bundle.Seed` isolation are retitled for what they test and now also assert the exact selected path.
- `selectLaunchDeployment` reports through an apierror constructor; message unchanged.

`go test ./cmd/...`, `-race -short ./cmd/app/...`, lint 0 issues, gofmt clean.

### DIFF
--- a/boot/deps/hub/dependency_handler.go
+++ b/boot/deps/hub/dependency_handler.go
@@ -1641,7 +1641,10 @@ func (h *DependencyHandler) refreshReplacementModuleIdentities(modules []Resolve
 		}
 		digest, size, err := digestReplacementTree(replacement)
 		if err != nil {
-			return NewDependencyIntegrityError(modKey(*mod), err, mod.Digest, mod.SizeBytes)
+			return apierror.New(apierror.Invalid, fmt.Sprintf("load local replacement %s from %q", modKey(*mod), replacement)).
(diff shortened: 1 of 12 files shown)


skhaz (approved):

wolfy-j:

PR #788, opened 2026-09-19
fix(supervisor): the first shutdown request decides the exit code

wolfy-j:
## Problem

`TriggerShutdown` called `setExitCode(code)` on every call. Only the SIGTERM was gated by `shutdownSent`, so the last caller's code won.

`service/terminal/host.go` `OnComplete` calls `TriggerShutdown` for every process that completes on the terminal host. When the launched command exits with 1, the SIGTERM stops any service sharing that host. That service's completion maps to 0 and overwrites the exit code.

Observed: `wippy test` in wippyai/framework `src/relay/test` exits 0 with a failing test, because `wippy.relay:central.service` runs on the same terminal host as the test runner. CI cannot see relay failures.

## Fix

The exit code is set only by the call that wins the CAS, which matches the function's own "only the first call" contract. The doc comment is updated.

Every caller was checked: the terminal host `OnComplete`, the `launchExecProcess` bridge (it passes the same code as the host), and `system` module exit. None relies on a later call changing the code.

## Tests

- New subtest `TriggerShutdown_FirstCallWinsExitCode`: calls with 1, then with 0, and asserts the exit code is 1 and exactly one signal is sent. It fails before the fix (`expected: 1, actual: 0`) and passes after.
- `go test ./api/supervisor/... ./service/terminal/... ./cmd/wippy/cmd/...` passes.
- golangci-lint (the repo's flags) reports 0 issues on the touched packages.


wolfy-j:
Closing: not needed. The failure came from wippyai/framework's relay test app placing its services and a spawned hub on the single-command terminal host. Fixed there in wippyai/framework#119; failing relay suites now exit non-zero on the current runtime.

PR #794, opened 2026-09-20
fix(naming): bind Strong outcomes to durable attempt identity

wolfy-j:
Strong registration can rewrite its pending record during node removal, and promotion gives the active record a different Raft epoch. Routing results by name, owner, or either entry epoch can therefore complete the wrong registration or leave its caller waiting. This change assigns each Strong registration a stable attempt ID, preserves it through pending rewrites and promotion, and delivers active/terminal outcomes only to that attempt. The returned active epoch remains the active record's Raft fencing index.

Pending deletion during promotion no longer releases the promoted exclusion. An older terminal event completes its own waiter without clearing a replacement. Terminal details are retained only while a caller waits. Existing active records without an attempt ID still load and release on deletion; newly created pending records require an ID.

Depends on #793 (which depends on #792). This PR does not change timer ownership after delayed writes or establish follower snapshot freshness; those remain separate correctness slices. The existing same-owner pending conflict response is also unchanged.

Validation: `go test -race ./system/topology/namereg/kvbacked -count=1`; real three-node follower promotion and leader-kill promotion under `-race`; `go vet ./system/topology/namereg/kvbacked ./cluster/clustertest`; `git diff --check`. Tests cover same-PID replacement, pending rewrite, promotion/delete, canceled waiter retention, legacy active startup/deletion, and returned active fencing epoch.


skhaz (approved):

wolfy-j:
Review found an upgrade/restart compatibility gap at head 5462d02745492fb096f7e09bcdd19f4f5b6a9354. I reproduced this with a focused Go test that writes the pre-change pending payload (p/n/d/r/dl, without a), then starts the reconciler: startup fails with registry record "_sys:registry:pending:claim": missing Strong attempt identity. Existing active records are covered, but existing pending records are not. This can strand startup after an upgrade with an in-flight claim; mixed-version writers also still create the old format. Holding merge pending an explicit safe transition/upgrade policy and regression coverage. Do not silently delete pending reservations or infer that old claims are absent.

wolfy-j:
Addressed the previously reported legacy-pending startup blocker. The patch does not rewrite pending entries or vote keys: a deterministic correlation token derived from the committed legacy entry is carried into promotion and recovered from deletion Previous records. Existing ACKs also restore local exclusions. Added race-tested legacy lifecycle regressions and real-Raft snapshot/durable-log recovery coverage; full affected suites, lint and build pass. The PR description now documents the upgrade contract and explicitly excludes mixed-version safety and the separate vote-format transitions.

wolfy-j:
Superseding the earlier compatibility approach per the alpha policy: removed the legacy token generator, pending compatibility decoder, old-active cleanup branch, and compatibility-only tests. Pending and Strong active records now strictly require attempt IDs and reject unsupported records explicitly. Current-format rejection tests, affected race suites, cluster proofs, lint and build pass. #795 is rebased without restoring the removed fallback.

PR #796, opened 2026-09-20
feat(naming): retain a versioned participant inventory without activation

wolfy-j:
Strong name claims currently choose participants from live gossip membership. A later change needs a durable, incarnation-qualified participant set before it can stop shrinking ACK requirements when a node temporarily disappears. This PR adds the inert persistence foundation: one bounded, versioned KV record with a stable domain, exact-revision CAS transitions, canonical node slots, per-node boot sequence/nonce, and retained retirement high-water marks.

An active incarnation cannot be replaced without explicit retirement. A new node begins at sequence 1; re-enrollment uses exactly the retired high-water plus one. Stale revisions and unknown or malformed records fail closed. A follower-local read is explicitly *observed*, not authoritative; after an uncertain response, callers must inspect the record and cannot silently retry against a newer revision. The format is documented separately from mesh frames, with a golden byte fixture for future independent implementations.

This package has no boot, gossip, Strong registration or admission wiring. Activating it requires a separately reviewed leader-fenced snapshot and participant lifecycle; recording a transition grants no naming authority.

Validation: `go test -race ./system/topology/namereg/... -count=1`; real three-node follower write, leader failover, post-failover write and durable restart under `-race` (including five repeated prepublication runs); scoped `golangci-lint` with the CI build tags; `go vet ./system/topology/namereg/inventory ./cluster/clustertest`; `git diff --check`. Tests include a competing commit between observation and CAS, an applied write with lost reply, corruption bounds, identity fencing, a 100-node record, and generation overflow.


skhaz (approved):

wolfy-j:
Reviewed for the small/minimal-surface merge batch. This PR explicitly adds an inert inventory package with no production boot, gossip, Strong or admission wiring. Holding it until a concrete reviewed consumer justifies this vocabulary and retained format; approval alone does not make unused infrastructure part of this batch.

wolfy-j:
Rebased on current `main` at `6daf554e3c`. The inventory remains inert: it does not enroll a node at boot or grant naming admission. This is one supported record format only; the magic/version discriminator rejects unknown bytes, with no compatibility fallback or mixed-version path.

Post-rebase validation passed: `go test -race ./system/topology/namereg/inventory/... ./cluster/clustertest/... -count=1`, scoped CI-tag `golangci-lint`, `go vet`, and `git diff --check`. Full PR CI is running.

PR #799, opened 2026-09-20
feat(kv): read bounded authoritative snapshots after a leader barrier

wolfy-j:
Follower-local KV snapshots may lag an accepted leader write. Naming needs a way to read selected keys, including authoritative absence, from one KV publication after a leader barrier. This adds a generic `AuthoritySnapshotReader` to the Raft-backed KV engine; followers forward over the existing relay and leaders return one detached immutable snapshot and KV publication revision.

The request and response formats are versioned and bounded (32 keys, 64 KiB), use connection-derived ingress identity from #798 for reply correlation, and preserve the relay package ownership contract. Reads and replies share bounded admission and operation deadlines; canceled callers remove pending waiters, while a noncancelable Raft barrier retains its permit until the future actually completes. Leadership and transport failures remain operational errors rather than malformed-input errors. No naming policy or workspace behavior is added here.

This PR is stacked on #798 (`fix/internode-source-provenance`) for the non-wire ingress provenance. #797 separately fixes standalone publication revisions before a later naming freshness-floor integration. It does not activate Strong naming reconciliation or assert production readiness.

Validation: full `go test -race ./system/kv ./cluster/clustertest -count=1`, focused admission/ownership/cancellation and real three-node partition tests, scoped CI-tag `golangci-lint`, `go vet`, and `git diff --check`. The three-node test uses real Raft with **in-process relay and synthetic ingress**, not production internode/TLS. A `GOARCH=386` package test is currently blocked by an unchanged `system/crdt/delta.go` constant-overflow build error; wire length validation checks unsigned bounds before converting to `int`.


skhaz (approved):

wolfy-j:
Batch architecture review: inspected the codec, admission, cancellation, ingress binding and forwarding paths; focused authority/relay race tests pass. No production caller of ReadAuthoritySnapshot exists in this PR: only the interface, implementation and tests. Under the current requirement to avoid unused runtime vocabulary, hold this protocol/API until a concrete consumer and its consistency contract land together. Also the description says 32 keys while maxAuthoritySnapshotKeys is 256; maxAuthorityConcurrent is 32. Current main now has LocalSnapshotReader in api/store/kv/snapshot.go, so integrate deliberately rather than replacing that file.

PR #803, opened 2026-09-20
test(cluster): prove Strong lifecycle across three TLS processes

wolfy-j:
Strong cluster naming had unit and in-process Raft coverage, but no test that independently booted processes could form the authenticated mesh, activate a Strong claim, observe its revocation on every replica, and shut down cleanly.

This adds a three-process test harness using production boot components, encrypted memberlist traffic, pinned node identities and mutual TLS for internode traffic. The parent verifies all three voters and one agreed leader, requires an active nonzero-epoch registration and observed revocation on each replica, and fails on child errors, missing shutdown acknowledgements or forced termination. Startup and teardown are bounded and joined; per-platform process control keeps Windows compilation working.

The socket proof is intentionally skipped under `go test -short`. Run it on a host permitting local TCP/UDP and Unix sockets:

```sh
go test ./cluster/processproof -run '^TestThreeProcessStrongClaimAndRevoke$' -count=1 -timeout=90s -v
go test -race ./cluster/processproof -run 'Test(ThreeProcessStrongClaimAndRevoke|HelperEarlyExitIsReported|HelperCleanExitWithoutShutdownIsReported)$' -count=1 -timeout=150s -v
```

Both passed locally on Linux. `go vet ./cluster/processproof`, `GOOS=windows GOARCH=amd64 go test -c`, and `git diff --check` passed. This is a lifecycle/connectivity proof; fault injection, authority-read barriers, sustained throughput and 100-node capacity remain separate gates.


skhaz (approved):

skhaz (approved):

wolfy-j:
Post-approval polish: removed the extra README, retained the run command in test source, and verified the real three-process TLS proof under -race plus Windows test compilation. The subsequent Ubuntu CI run exposed TestProxyRunsNativeInteractiveEditingKeys sending successive edits while Bash was between readline sessions. Reproduced locally in 100 race runs; the fixture now waits for each prompt and has a bounded cancellation context. All editing assertions remain. After the fix: 100 focused race runs pass, full terminal-proxy and process-proof race suites pass, and scoped lint reports zero issues. No production behavior changes; no new review requested. Waiting for the rerun of hosted CI before merge.

PR #806, opened 2026-09-20
fix(naming): keep Strong votes bound to one attempt across rewrites

wolfy-j:
When a required node leaves during a Strong reservation, the leader rewrites the pending header to update its participant set. KV assigns that write a new Raft epoch, so vote keys derived from the pending entry's epoch no longer match the surviving nodes' ACKs. A claim can then remain pending despite having all remaining votes.

Give each new Strong reservation an opaque 128-bit attempt ID before submitting its pending transaction. Carry that ID in pending and active records, and use it for ACK/reject keys and cleanup. The entry epoch remains the Raft publication fence and returned epoch; a header rewrite no longer changes vote identity. Reject malformed pending IDs during seed instead of allowing unrelated attempts to share a vote namespace.

The real three-node Raft regression holds one participant back, waits for two committed ACKs, rewrites the pending record at a new epoch, and proves that the existing votes still allow promotion. It passed five consecutive runs under `-race`. Full KV-backed naming and cluster race suites, focused lint and vet also pass locally.

Stacked on #805. This fixes vote identity; the separate reconciler-owner work will address the caller's post-write read, historical outcome delivery, stale timers, and the all-pending scan. It does not change the participant-fencing policy.

Deployment note: pending records written by an older binary do not contain an attempt ID and will fail closed during seed. Drain in-flight Strong reservations before deploying this storage-format change; mixed-version Strong writers cannot safely share the vote namespace. No compatibility or migration machinery is added without a verified consumer need.


skhaz (approved):

wolfy-j:
Batch review: reproduced the documented upgrade incompatibility at 5ea1445. A pending payload without codec field a (the format actually shipped in v0.3.42a) makes StartReconciler fail with invalid Strong attempt identity. This is not merely a hypothetical mixed development stack. Holding merge under the existing no-surprises upgrade requirement until we choose and test either a coordinated drain/upgrade contract or a safe migration. This also overlaps #794/#800 attempt identity; do not land a second divergent identity model. No existing reservation should be silently deleted or assigned an invented identity.

wolfy-j:
Superseded by main: attempt-bound Strong votes landed in #794 (527694c884) and #800 (06e6213a04, aa71abfd16); the rewrite scenario is covered by TestStrongMembershipRewriteKeepsAttemptEvidence and the three-node real_raft test.

PR #807, opened 2026-09-20
fix(naming): route Strong vote events directly with unambiguous keys

wolfy-j:
Strong ACK/NACK watch events currently scan and reconcile every pending claim. With many unrelated claims, each vote repeats work across the entire pending keyspace. This change routes an event to its exact pending attempt and reconciles that one name; unknown or stale votes do not drive another attempt.

Vote keys now encode the UTF-8 **byte length** of the name before the name, attempt ID, and node ID. The earlier colon-separated format is ambiguous when names and node IDs contain colons. The parser checks canonical bounds, the attempt identity, and required-node membership against the local pending header. There is no name index, full pending scan, or additional leader read on the vote path. General cluster events remain on the event bus; KV key updates use the KV-owned watch.

**Rollout constraint:** this PR is stacked on #806 and changes stored vote keys. Do not run old and new format participants together. Drain old pending Strong attempts and inspect/remove orphan old-format vote keys before deploying the new format. This PR does not implement mixed-version migration. It also does not resolve the separate Strong caller/reconciler race or participant fencing policy.

Validation: full `system/topology/namereg/kvbacked` package under `go test -race`; focused real cluster integration test; `go vet`; `git diff --check`. Regressions cover zero vote-triggered pending scans, NACK progression, names/node IDs containing colons and Unicode, empty names, stale attempts, malformed/oversized lengths, and late vote deletions. The earlier synthetic benchmark showed why fanout matters but predates #805/#806, so no speedup percentage is claimed for this branch.


skhaz (approved):

skhaz (approved):

wolfy-j:
Batch review: new-format routing is focused and avoids the full pending scan, but the rollout boundary is real. A focused Go reproduction writes an ACK in the parent #806 format and calls complete for the same name/attempt/required node: the committed ACK is no longer recognized. Holding with #806 until the supported upgrade path is chosen and exercised. Preserve the unambiguous byte-length grammar, but do not add ad hoc dual-format routing without deciding how mixed writers and existing claims are fenced.

wolfy-j:
Review against the current main/#800 sequence: direct vote routing is useful, but this branch must not reintroduce #806’s older state-machine implementation. strongVoteName also currently treats every engine.Get/decodePending error as an unrelated vote; non-not-found storage and malformed current-record errors must propagate to the reconciler’s existing fail-closed path. Keep one canonical vote grammar when rebasing, and retain #794/#795 strict identity/timer behavior and #808 discovery-loss semantics. Holding this PR until that integration and error-path coverage are verified.

PR #810, opened 2026-09-20
Expose coherent local KV publication revisions

wolfy-j:
Strong name admission needs to compare a local key snapshot with later KV watch events. Separate key reads can observe different transactions, and the standalone backend previously published a delete-only snapshot without advancing its revision. This change gives local reads, prefix scans, and watch events a comparable publication revision while preserving the existing Raft index used as the deletion/dissemination dot.

`LocalSnapshotReader` and `LocalSnapshotScanner` return detached data from one immutable publication without a network round trip. All events from a transaction carry that publication's revision. Actual deletions advance the standalone state's version; unsuccessful deletions do not. The general event bus remains for topology hints; key observations remain in the KV-owned watch.

Validation: `go test -race ./system/kv ./api/store/kv -count=1`, transaction watch ordering and delete-only publication regressions, and `git diff --check`.

Stacked on #809. Since the KV state version sequence changes on deletes, deploy matching binaries across Raft voters; mixed-version rollout is outside this PR's contract.


skhaz (approved):

wolfy-j:
Integration gate after merging #797: this branch currently calls nextVersion() in state.del(), whereas #797 deliberately advances a separate publication revision to preserve persisted Entry.Version allocation and replay of already-committed CAS commands. Rebase/reconcile this change onto main without restoring deletion-driven entry-version increments; retain and run the full-log and snapshot-plus-tail replay regressions added by #797. Current approval/green functional tests do not remove this compatibility requirement.

PR #811, opened 2026-09-20
Serialize Strong observation and per-name admission

wolfy-j:
A required node could commit a Strong ACK before excluding competing LOCAL/EVENTUAL names. A later watch event for an older active owner could then replace or remove a newer pending exclusion. This change serializes each name's local admission decision with Strong exclusion, and makes the ordered KV watch own the attempt's transitions. A worker may perform a conditional vote or retry, but it cannot relatch a retired attempt after the watch advances.

The reconciler retains state proportional to live obligations while bounding concurrent KV actions with a fixed worker pool. It seeds active and pending claims from one immutable local publication, enrolls result waiters before writing pending, keeps timers and terminal results attempt-bound, and validates promotion against the committed vote set. LOCAL and EVENTUAL share the per-name admission coordinator. KV watches carry key transitions; the general event bus carries node/link hints, never naming authority.

On a nonmember client, ordinary name resolution can still forward. Fresh LOCAL/EVENTUAL/Strong admission is deliberately unavailable because an active-key miss cannot prove absence of a pending Strong claim; an authoritative pending-state feed is required to enable it safely. This is a visible behavior change for apps that create local names on cluster clients.

Validation: affected KV/naming/topology race suites, 30 repetitions of unregister and stale-event regressions, boot tests with localhost binding, real three-node Strong/client/failover tests, `go vet`, and `git diff --check`. Astra reviewed the scoped ownership/admission ordering and found no remaining P1 blocker in that slice.

Stacked on #810 (itself on #809 and #808). **Draft for Rodrigo's review:** node activation/retirement is not yet authoritative, existing Strong attempts still derive required nodes from gossip, and 100-node toxic-network/load evidence is outstanding. This PR does not claim full cluster readiness. KV delete-version changes require matching voter binaries.


skhaz (approved):

wolfy-j:
Re-reviewed against current `main` with Astra. Keeping this PR draft; the published head is not merge-ready yet.

Blocking integration work:

1. Failed or malformed KV observations must close naming admission. The published head returns from Strong snapshot/decode failures without closing it and ignores malformed active/pending watch records. Preserve the current fail-closed reconciliation behavior in the ordered-owner model.
2. One Strong attempt must never commit both ACK and NACK. Check both vote keys in the transaction and retain the `TestStrongAckAndNackCannotBothCommit` regression.
3. Preserve the current cross-scope `LookupOther(name, proposedPID)` behavior, escaped vote-key grammar, reconnect cleanup, and attempt-bound timer protections. A semantic rebase is required; resolving text conflicts alone is insufficient.
4. Port the current regression tests to the new ordered-owner and committed-terminal-event APIs without dropping their safety assertions. An isolated integration attempt still fails test compilation on obsolete waiter/terminal helpers and a malformed vote-route test fixture.

This is separate from #796 (inert participant inventory) and #799 (authority snapshot reader). Neither by itself activates participant enrollment or grants ongoing naming admission.

wolfy-j:
Integration update (main after #796/#799): I split the missing coherent local prefix scanner into #829. #811 still conflicts with main and remains draft. The rebase must preserve main’s escaped vote-key grammar, proposed-PID cross-scope checks, local-only EVENTUAL revocation, and authority/snapshot behavior. Before marking it ready, I will also require fail-closed observation errors and an atomic check that ACK and NACK cannot both be committed for one node/attempt. The initial automatic-merge trial does not compile and has not been pushed. Activation/retirement authority and 100-node evidence remain separate cluster-readiness work as already noted in this PR.

wolfy-j:
Prerequisite #829 is now merged on main (a060c74265). The remaining #811 integration is the naming owner/admission change itself; this draft is not merge-ready yet.

skhaz (approved):

wolfy-j:
Adversarial follow-up (def75cd50): Strong voting now inspects every live EVENTUAL dot for the name, not just the visible winner. EVENTUAL lookup suppresses a conflicting result while this node holds a pending or active Strong exclusion; the raw CRDT state remains intact for convergence. Regressions cover hidden local and remote dots, delayed gossip after reservation, and a real KV Strong NACK. Affected race suites, 20-run targeted stress, and go vet pass locally. Participant retirement is still deliberately fail-closed and remains a separate design/rollout decision; a gossip timeout is not fencing.

wolfy-j:
Integration update (57605ea97): merged current main, including #830, without text conflicts. Affected naming/boot/real-cluster race suites, topology/membership tests, and the separate-process cluster proof pass on the combined tree. Added a regression assertion that Strong voting ignores a superseded prior-incarnation EVENTUAL dot. Fresh CI is running. Keeping draft while the safe participant-retirement/fencing contract and coordinated rollout are resolved; a gossip timeout is not a fence.

wolfy-j:
@skhaz The current head 57605ea97 incorporates #830 and all nine CI checks now pass. I requested a fresh review because the prior approval predates the integration. The protocol remains draft pending the explicit node-fencing/retirement and rollout decision; please review the integrated naming semantics meanwhile.

skhaz (approved):

skhaz (approved):

